PR #6, opened 2025-04-30
Persistence model notebook

npedrazzini:


IFenton:
At the moment this is crashing for me on the Dataset Creation step (cell 8), as it's still looking for an ERA5 variable `uas`:
Screenshot 2025-05-02 at 12 23 08

marianovitasari20:
If I’m not mistaken, that error was raised because it couldn’t detect the dataset. That’s why it couldn’t detect the variable either. Have you tried using the prototype datasets in the notebook, @IFenton?  If yes and it's still not working, maybe either you or @npedrazzini could try running it using Nilo's environment (short version) and the prototype datasets I provided in the shared folder (`/.../vjgo8416-ice-frcst/shared/prototype_data`) to see if you get the same error.

If the error still persists, I think it’s also worth checking whether creating a new environment based on the full version of the YAML file I shared resolves the issue. I remember that back then, I couldn’t use their environment from IceNet repo to run the notebook—after debugging here and there, creating a new environment (with the version I have) helped solve the problem.

So far, I haven’t had any errors with my environment, but I haven’t tried running the full script with Nilo’s short version env. I just ran it to make sure there were no import errors.

IFenton:
@marianovitasari20 I should have made it clearer, this error relates specifically to the Persistence model notebook which shouldn't be using ERA5. I didn't check the other notebooks here, though I suspect your solution would work for them

IFenton:
This is currently giving an error, when run on Baskerville, which I am looking in to. It runs fine locally
Screenshot 2025-05-16 at 09 25 19

Screenshot 2025-05-16 at 09 24 07

IFenton:
That bug can be fixed by adding in the following code:

```
import os 

# Unset SLURM_NTASKS if it's causing issues
if "SLURM_NTASKS" in os.environ:
    del os.environ["SLURM_NTASKS"]

# Optionally, set SLURM_NTASKS_PER_NODE if needed
os.environ["SLURM_NTASKS_PER_NODE"] = "1"  # or whatever value is appropriate
```

marianovitasari20:
Yes, those lines are the ones I recently added two weeks ago during our in-person meeting @IFenton (previously, I had excluded committing those lines).

IFenton:
I think this is good to go after the original notebooks have been merged in

IFenton (approved):
LGTM

PR #8, opened 2025-06-24
Add initial Anemoi support

jemrobinson:
## Typer entrypoint

Replaces the `argparse` entrypoint with a `Typer` entrypoint. Run `uv run zebra --help` to see available commands. Note that most commands take a `--config-name` option to override the default Hydra config and/or arbitrary space-separated Hydra overrides like `+my.option.here=5 +my.option.there='zebra'`

## Anemoi dataset

Config files in `config/`. Dataset loaders in `datasets/`. For SIC, I've added a slightly-hacky pre-processing step which runs the `SICDownloader`. Ideally, we'd register a custom Anemoi class for this, but I thought it was more important to get something working first.


ice_station_zebra/cli/__init__.py
```diff
@@ -0,0 +1,7 @@
+from .hydra import hydra_adaptor
```
louisavz, line 1: Recommend to not use relative import to maintain maximum flexibility and ease of maintainability if we need to restructure the code.

ice_station_zebra/cli/hydra.py
```diff
@@ -0,0 +1,56 @@
+import inspect
+import itertools
+from typing import Annotated, Any
+
+from hydra import compose, initialize
+from omegaconf import DictConfig
+from typer import Argument, Option
+
+
+def hydra_adaptor(function):
```
louisavz, line 10: Suggest adding return type for readability and consistency with your other functions.

ice_station_zebra/cli/__init__.py
jemrobinson: I'm not sure how to avoid that here. This is the file that defines what is available when you run `from ice_station_zebra.cli import X`. Is there a way to use non-relative imports here?

ice_station_zebra/cli/hydra.py
jemrobinson: Done in 190144a

ice_station_zebra/cli/hydra.py
```diff
@@ -0,0 +1,56 @@
...
+    """Replace a function that takes a Hydra config with one that takes string arguments
+
+    Args:
+        function: Callable(*args, config: DictConfig, **kwargs)
+
+    Returns:
+        Callable(*args, config_name: str, **kwargs, overrides: list[str])
+    """
+
+    def wrapper(
```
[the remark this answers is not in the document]
jemrobinson: Done in 5569c5a

jemrobinson:
Probably easiest to debug in person, but do you actually have a folder called `/data` at the root level of your filesystem?

PR #32, opened 2025-07-22
Improve model checkpointing

jemrobinson:
- Move anemoi datasets and lightning dataloaders into a common folder
- Keep model checkpoints in `wandb` run folder (with both under $base_path / training)


IFenton (approved):
LGTM

PR #54, opened 2025-08-13
Add a persistence model

### LINKED ISSUE #22: Integrate persistence model into pipeline
## Problem / Goal
The persistence model code lives in a notebook and cannot be called by the main pipeline.
We aim to integrate the persistence model with the broader pipeline. This will help set the standard for how to integrate future models that are still in dev.

## What does good look like?
- [x] A model class exists in ice_station_zebra/models/model-name.py, implemented as a pytorch_lightning.LightningModule with required methods (forward, training_step, validation_step).
- [x] Default hyperparameters for the diffusion model are defined in ice_station_zebra/config/model/model-name.yaml and are accessible via Hydra.
- [ ] Documentation for the diffusion model, including configuration and usage, is present in docs

jemrobinson:
Add a basic persistence model that does the following:

- predicts the next `n_forecast_steps` of output by duplicating the last input step N times
- turns off optimisation during training (as there's nothing to optimise)
- adds a new checkpoint callback to save a checkpoint even if no optimisation has happened

Also adds a couple of small fixes to the SIC plotting code

Closes #22

### DIFF
--- a/ice_station_zebra/callbacks/unconditional_checkpoint.py
+++ b/ice_station_zebra/callbacks/unconditional_checkpoint.py
@@ -1,3 +1,5 @@
+from pathlib import Path
+
 from lightning import LightningModule, Trainer
 from lightning.pytorch.callbacks import Callback, ModelCheckpoint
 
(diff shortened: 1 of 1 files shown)


IFenton (approved):
LGTM. The only issue seems to be in the return type of dirpath, which is causing the mypy test to fail

PR #60, opened 2025-08-15
Make activation function configurable as argument

marianovitasari20:
Previously, activation functions were hardcoded in the blocks. 
Now they can be set via the configuration, allowing different models to use different activations without changing code.


jemrobinson (changes requested):
Could we change the type-signature of these modules to take `str` instead of the function itself? This makes it easier to serialise for checkpointing.

PR #74, opened 2025-08-21
Add EMA callback

jemrobinson:
@marianovitasari20 @erinuclkwon this is the callback that will be included in the next version of `pytorch-lightning`. I've added a small wrapper class that lets you set `every_n_epochs` and `every_n_steps` which control how often the EMA averaging will be applying.

I've added this as one of our default callbacks for training, so it will be applied to every model by default. Does that do everything you need?


erinuclkwon (approved):
Yes, it does. LGTM. Thanks!

PR #80, opened 2025-08-27
Fix persistence model

### LINKED ISSUE #65: Checkpoint creation
### Problem / Goal

In the last couple of days, the evaluate command has started producing two checkpoint files at the end of a run, e.g. `epoch=49-step=700.ckpt` & `epoch=50-step=700.ckpt`. Prior to that only one file was produced, e.g.. `epoch=50-step=0.ckpt`. I think this is since we merged in the persistence model (and added the `unconditional_checkpoint.py` file). 

We need to decide whether this is a feature we want, or whether it is a bug (and if it's the latter, then we should fix it).

### What does good look like?

 - [ ] Decide whether this is a bug that needs fixing
 - [ ] If so, implement a fix

### Related context

_No response_

### LINKED ISSUE #78: Fix issue with persistence model
### Problem / Goal

The persistence model is no longer running. I think this is related to the PR that merged in the `EMAWeightAveragingCallback`. I get the following error: `RuntimeError: Expected !nested_tensorlist[0].empty() to be true, but got false.  (Could this error message be improved?  If so, please report an enhancement request to PyTorch.)`

### What does good look like?

- [ ] Persistence runs

### Related context

_No response_

jemrobinson:
Fix the persistence model by updating the configs:

- Set `unconditional_checkpoint` to only run for persistence
- Set `ema_weight_averaging` to only run for *not-persistence*
- Add a top-level `persistence.yaml` config to demonstrate how to enable these options

Closes #78. Closes #65.


IFenton (approved):
LGTM

PR #87, opened 2025-08-29
Process in NTCHW space

### LINKED ISSUE #77: Use NTCHW tensors in processing
### Problem / Goal

Currently our encode-process-decode architecture squashes time x channels into one dimension for the processor (i.e. we go from an NTCHW tensor to a NCHW tensor).

This was mainly done so that processors didn't have to worry about taking N input timesteps and producing M (!=N) output timesteps.

However, several types of model (e.g. transformers) might like the time dimension while others (e.g. UNet) can do an internal rollout to produce multiple outputs.

### What does good look like?

- [ ] Replace NCHW with NTCHW in processing

### Related context

_No response_

jemrobinson:
Update the encode-process-decode structure as follows:

**Encoder:**
input: NTCHW (batch_size, n_history_steps, input_channels, input_height, input_width)
output: NTCHW (batch_size, n_history_steps, latent_channels, latent_height, latent_width)

**Processor:**
input: NTCHW (batch_size, n_history_steps, n_latent_channels, latent_height, latent_width)
output: NTCHW (batch_size, n_forecast_steps, n_latent_channels, latent_height, latent_width)

**Decoder:**
input: NTCHW (batch_size, n_forecast_steps, latent_channels, latent_height, latent_width)
output: NTCHW (batch_size, n_forecast_steps, output_channels, output_height, output_width)

Add a default `rollout` structure to `BaseProcessor` which means that models can just implement `rollout` to perform a single `NCHW -> NCHW` operation, and this will be applied repeatedly to produce `n_forecast_steps` outputs for `forward`.

Closes #77

### DIFF
--- a/ice_station_zebra/models/decoders/base_decoder.py
+++ b/ice_station_zebra/models/decoders/base_decoder.py
@@ -34,9 +34,8 @@ class BaseDecoder(nn.Module):
         """
         return stack(
             [
-                self.rollout(
-                    x[:, idx_t, :, :, :]
-                )  # cut the NTCHW input into NCHW slices
+                # Apply rollout to each NCHW slice in the NTCHW input
+                self.rollout(x[:, idx_t, :, :, :])
                 for idx_t in range(self.n_forecast_steps)
             ],
             dim=1,
--- a/ice_station_zebra/models/encoders/base_encoder.py
+++ b/ice_station_zebra/models/encoders/base_encoder.py
@@ -34,9 +34,8 @@ class BaseEncoder(nn.Module):
         """
         return stack(
             [
-                self.rollout(
-                    x[:, idx_t, :, :, :]
-                )  # cut the NTCHW input into NCHW slices
+                # Apply rollout to each NCHW slice in the NTCHW input
+                self.rollout(x[:, idx_t, :, :, :])
                 for idx_t in range(self.n_history_steps)
             ],
             dim=1,


IFenton:
Reviewing this

ice_station_zebra/models/encode_process_decode.py
```diff
@@ -9,7 +9,9 @@
 from .zebra_model import ZebraModel
 
 if TYPE_CHECKING:
+    from ice_station_zebra.models.decoders import BaseDecoder
     from ice_station_zebra.models.encoders import BaseEncoder
+    from ice_station_zebra.models.processors import BaseProcessor
 
 
 class EncodeProcessDecode(ZebraModel):
```
IFenton, line 17: Do you also need to update the forward method in this class? I think stuff like the self.processor now uses NTCHW?

ice_station_zebra/models/processors/null.py
```diff
@@ -1,29 +1,40 @@
...
+        Args:
+            kwargs: Arguments to BaseProcessor.
+
+        """
+        super().__init__(**kwargs)
         self.model = nn.Identity()
 
-    def forward(self, x: TensorNCHW) -> TensorNCHW:
-        """Forward step: process in latent space.
+    def rollout(self, x: TensorNTCHW) -> TensorNTCHW:
```
IFenton, line 30:
```suggestion
    def rollout(self, x: TensorNCHW) -> TensorNCHW:
```
I think?

IFenton (changes requested):

ice_station_zebra/models/encode_process_decode.py
jemrobinson: Thanks - this has helped me find quite a few bugs!

1. we need to concatenate on dimension 2 now (fixed in 10479d2)
2. predictions are a valid input (fixed in 25902b8)
3. decoders need to know the total number of latent channels (fixed in e429a43)

ice_station_zebra/models/processors/base_processor.py
```diff
@@ -0,0 +1,67 @@
...
+            nchw_slices.append(outputs[-1])
+
+        # Stack the outputs up as a new time dimension
+        return stack(outputs, dim=1)
+
+    def rollout(self, x: TensorNCHW) -> TensorNCHW:
+        """Single rollout step: process in NCHW latent space.
+
+        Args:
+            x: TensorNTCHW with (batch_size, n_history_steps, n_latent_channels, latent_height, latent_width)
```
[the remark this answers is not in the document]
jemrobinson: Done in 48c2762

PR #89, opened 2025-09-05
Make common blocks more flexible

### LINKED ISSUE #68: Refactor common blocks for flexibility
### Problem / Goal

The current common block can only be used for the current U-Net version and is not flexible enough to be used for other models (e.g., DDPM). The goal is to modify the common blocks (ConvBlock, UpConvBlock, and BottleneckBlock) to accommodate other models and make them more flexible.

### What does good look like?

- [ ] The common blocks can be used for both U-Net and DDPM architectures.
- [ ] Flexible options for normalization, dropout, etc.
- [ ] The changes are tested and verified to ensure consistency.

### Related context

https://github.com/alan-turing-institute/ice-station-zebra/tree/flexible-common-block

marianovitasari20:
Closes #68


ice_station_zebra/models/common/convblock.py
```diff
@@ -4,46 +4,129 @@
 
 
 class ConvBlock(nn.Module):
```
jemrobinson, line 6: I think there could be 4 (or 8?) different classes here for the different combinations of `normalization`, `activation_after_norm` and `final`. Are all of these useful? Can we create some explicit, less-flexible classes for just the ones we want to use?

It's much harder to see where two ConvBlocks with different options are similar/different than if one of them is e.g. ConvBlockBatchNorm and one is ConvBlockGroupNorm.

jemrobinson:
:eyes: Reviewing this now

PR #104, opened 2025-09-24
E2E Tests for Plotting

LydiaFrance:
As our visualisations will influence hyper-parametrising and other refinement checks for people running models, I wanted to introduce tests which will check the plots and videos are being produced as expected. Testing code end-to-end means we are checking the behaviour on a known test case does what we expect. 

To do this there are two ways the tests run:

### Fake Data

This is a small synthetic array that is shaped a little like sea ice data [T, H, W] with 0-1 values. This means we can produce plots very quickly for testing. Catches shape mismatches or similar in the plotting code. 

### Example checkpoint

A minimal frozen `.ckpt` file has been saved inside tests. In future we would want to have end-to-end tests that include running a tiny version of the data through our models and afterwards test the plots with them, but for now this small file is helpful. There's nothing useful in this checkpoint, it's not from a real trained model. 

Confirms the full pipeline works: checkpoint -> model -> outputs -> callback -> figure objects (images/videos)

## Warning Badge

I added functionality which puts red text under the title of the plots if something is wrong with the colour scale or the scale of the prediction in comparison to the ground truth. 

### API behaviour

- plot_maps returns a valid PIL image.
- video_maps returns a valid MP4/GIF video buffer.
- These are the exact objects that callbacks log to wandb.

### Subplot Layout

No overlapping panels and colourbars for...

- 2 panel vs 3 panel figures 
- Horizontal vs vertical colourbars
- Shared vs separate colourbars

### Colour scaling

- Difference plots use a symmetric diverging scale centred on 0
- Tick marks are sensible
- Shared colourbars apply the same vmin/vmax to both Ground Truth and Prediction

### Range Check Reporting

- Warnings are correctly generated if values fall outside the displayed colour range, or if the Prediction is much higher/lower than the Ground Truth
- Confirms a warning badge is displayed under the figure title

### Animations

- Uses monkeypatching to prevent matplotlib trying to open a window and slowing down the testing
- Doesn't require the CI environment to have `ffmpeg`

## Fixtures

In python, pytest fixtures prepare some data or objects that exist during the test, they act as bioler plate code the tests can call when needed. 

- Fake data fixture
- ModelTestOutput from the tiny checkpoint file
- PlotSpec 
Here’s a concise, researcher-friendly “how to run the tests” section you could drop into your PR or your README. It’s minimal, practical, and shows the exact commands.

⸻

## Running the tests with Pytest

We use [pytest](https://docs.pytest.org/) for testing. 

```terminal
pytest
```

By default this will also collect coverage information (because of our config). If you don’t care about coverage output, add `--no-cov`:

```terminal
pytest --no-cov
```
If you just want to check the new visualisation code:

```terminal
pytest tests/plotting --no-cov
```

For example, only test the plotting API:

```terminal
pytest tests/plotting/test_api.py --no-cov
```

### DIFF
--- a/tests/plotting/conftest.py
+++ b/tests/plotting/conftest.py
@@ -134,14 +134,16 @@ def checkpoint_data(
         # Load config (try checkpoint dir first, fallback to default)
         config_path = example_checkpoint_path.parent.parent / "model_config.yaml"
         if config_path.exists():
-            config = cast("DictConfig", OmegaConf.load(config_path))
+            config = OmegaConf.load(config_path)
+            assert isinstance(config, DictConfig)
         else:
             # Fallback to default config
             # Look for a file ending with local.yaml
             config_dir = Path("ice_station_zebra/config/")
             yaml_iter = config_dir.glob("*.local.yaml")
             local_yaml = next(yaml_iter)
-            config = cast("DictConfig", OmegaConf.load(local_yaml))
+            config = OmegaConf.load(local_yaml)
+            assert isinstance(config, DictConfig)
 
         # Load model from checkpoint
         model_dict = cast(


.pre-commit-config.yaml
```diff
@@ -24,7 +24,7 @@ repos:
       # Run the linter
       - id: ruff-check
         args: ["--fix", "--show-fixes"]
-        pass_filenames: false
+        # pass_filenames: false
```
aranas, line 27: Did you comment this out to avoid running tests on whole project during dev? If so, should this be re-enabled before merging?

ice_station_zebra/visualisations/plotting_maps.py
```diff
@@ -406,6 +447,31 @@ def _draw_frame(  # noqa: PLR0913
                 vmax=vmax,
             )
 
+    # Optional: visually mark NaNs as semi-transparent grey overlays
+
+    def _overlay_nans(ax: Axes, arr: np.ndarray) -> None:
```
aranas, line 452: Why was this function added if not currently used?

tests/plotting/test_dummy_arctic.py
```diff
@@ -0,0 +1,48 @@
+from __future__ import annotations
```
aranas, line 1: `test_dummy_arctic.py` looks more like a helper for generating fake sea ice data than an actual test file. The `test_ `prefix in the name is a bit misleading — maybe renaming it to something like create_dummy_arctic.py helper_dummy_data.py would make its purpose clearer?

tests/plotting/conftest.py
```diff
@@ -0,0 +1,234 @@
...
+
+from .test_dummy_arctic import make_circular_arctic
+
+mpl.use("Agg")
+
+TEST_DATE = date(2020, 1, 15)
+
+
+@pytest.fixture
+def sic_pair_2d(
```
aranas, line 23: Non-blocking comment: Would’ve expected sic_pair_3d_stream to be based on sic_pair_2d rather than the other way around. That said, changing it would probably not be worth the extra effort, so no action needed.

aranas (commented):
apart from that above question about the magnitude warning, this PR looks good to me. Have added some minor comments inline.

ice_station_zebra/visualisations/plotting_maps.py
LydiaFrance: Because our data isn't applying any masks currently... I am unsure why! But I wanted to leave the function there for the future.

tests/plotting/conftest.py
LydiaFrance: To get 2d from the data you start with a stream first anyway so that way around prevents repetitive code

.pre-commit-config.yaml
LydiaFrance: I had extra files I wasn't committing to the codebase so it was failing and I wanted it to just check what i was developing. I can remove it though!

LydiaFrance:
> Overall this looks great — thanks for putting this together @LydiaFrance . I just have one bigger question:
> 
> In `sanity.py`, the magnitude mismatch check (_robust_p2_p98) feels more like a model quality issue than a visualization concern. Am I reading this right? Would it make more sense to move this into the model testing suite, or is there a specific reason it’s useful to surface at the plotting stage?
> 
> It would also be good to run the tests in this PR against the latest checkpoints to make sure everything works as expected — I don’t have access to Baskerville. Have you already run this with a recent checkpoint? If not, maybe someone else can do that before we merge.
> 
> I’ll leave a few smaller comments inline, but I wanted to raise these points early.

Thanks Sophie! Appreciate your review :) 

It does seem like a model quality test but it is actually about what colourscale has been chosen by the user. So if we imagine the user selected the plots can have their own colourbar, and the prediction map looks identical to the ground truth, but in reality the prediction scale is 100x the ground truth. The colours will make the prediction look good, so this additional subtitle is flagging that for the user. And the other kind of warning is that by sharing a colour scale, the prediction is not being shown adequately by the plot (the numbers are outside the colour limits). We definitely need model quality tests but this is specifically a warning about colour scale choice.

LydiaFrance:
As requested, some images of what the figure looks like when there's a colour scale problem:
sea-ice_concentration-static-maps_0_cb77d946fdb2172519d1

LydiaFrance:
I have changed `dummy` to `fake data` in the code base, and changed `sanity` to `range check`.

tests/plotting/conftest.py
```diff
@@ -0,0 +1,259 @@
...
+    Returns (ground_truth, prediction, date) if checkpoint is available, else None.
+    """
+    if not use_checkpoint or example_checkpoint_path is None:
+        return None
+
+    try:
+        # Load config (try checkpoint dir first, fallback to default)
+        config_path = example_checkpoint_path.parent.parent / "model_config.yaml"
+        if config_path.exists():
+            config = cast("DictConfig", OmegaConf.load(config_path))
```
jemrobinson, line 137: This is fine but typically it's better to avoid `cast` where possible. Since we're in a test here, something like

```
config = OmegaConf.load(config_path)
assert isinstance(config, DictConfig)
```

would probably be better, as it makes the test fail if our assumption (that this is loading a DictConfig) is wrong.

PR #107, opened 2025-09-29
Better encoder decoder structure

jemrobinson:
- Move latent space specification to encoder block
- Allow number of channels in latent space to be determined by encoder rather than fixed
- Added resizing interpolation to replace Upsample
- Simplified names of data spaces passed to encoder, processor and decoder (now always called `data_space_in` and `data_space_out`)


tests/models/test_decoders.py
```diff
@@ -12,34 +12,34 @@
 class TestDecoders:
     @pytest.mark.parametrize("test_batch_size", [1, 2, 5])
     @pytest.mark.parametrize("test_decoder_cls", ["CNNDecoder", "NaiveLinearDecoder"])
-    @pytest.mark.parametrize("test_latent_shape", [(32, 32, 128), (200, 100, 3)])
+    @pytest.mark.parametrize("test_latent_chw", [(128, 32, 32), (2, 200, 100)])
```
louisavz, line 15: For my own learning - I've never used pytest.mark.parametrize and I'm not 100% sure what it is doing. I see you are defining test_latent_chw to be two sets of tuples. Given this is for the decoders, are you expecting the test_output_chw's dimensions to match given the corresponding test_latent_chw? Also you are testing two cases here?

tests/models/test_decoders.py
```diff
@@ -12,34 +12,34 @@
...
         test_decoder_cls: str,
-        test_latent_shape: tuple[int, int, int],
-        test_output_shape: tuple[int, int, int],
+        test_latent_chw: tuple[int, int, int],
         test_n_forecast_steps: int,
+        test_output_chw: tuple[int, int, int],
     ) -> None:
         latent_space = DataSpace(
-            name="latent", shape=test_latent_shape[0:2], channels=test_latent_shape[2]
+            name="latent", channels=test_latent_chw[0], shape=test_latent_chw[1:3]
```
louisavz, line 27: Could use test_latent_chw[1:] here and for below instances

tests/models/test_decoders.py
```diff
@@ -56,3 +56,20 @@ def test_forward_shape(
...
+
+
+class TestCNNDecoder:
+    def test_latent_shape_errors(self) -> None:
+        test_n_forecast_steps = 1
+        latent_space = DataSpace(name="latent", shape=(32, 32), channels=3)
+        output_space = DataSpace(name="output", shape=(256, 256), channels=4)
+        with pytest.raises(
+            ValueError,
+            match="The number of input channels 3 must be divisible by 2. Without this, it is not possible to apply 1 convolutions.",
```
louisavz, line 68: Apologies, a nitpick. The error message on "input channel / latent space channel" might benefit from not being hardcoded, but use the variable channels above (latent_space.channels).

louisavz (approved):
Did a visual review and all looks good to me. I have a question for my own learning and couple minor optional suggestions in the comments. Thanks for putting this PR together!

tests/models/test_decoders.py
jemrobinson: It just means "run the test once for each combination of each of these variables". So there are 3 batch sizes x 2 decoder_cls x 2 latent_shape x 2 latent_chw => 24 times the test gets run with different arguments.

tests/models/test_decoders.py
jemrobinson: Done in fa497a8

tests/models/test_decoders.py
jemrobinson: Done in cd6a4e8

PR #115, opened 2025-10-06
Improved cnn encoder decoder

jemrobinson:
Tested various potential improvements to the CNNEncoder and CNNDecoder. Many of these did not work at all, but the best performing (after 100 epochs of training with 1 year) is the following:

**Encoder**
- Resize with interpolation (if needed) so that the shape after convolution will *exactly* match the desired output shape
- n_layers of size-reducing convolutional blocks

**Decoder**
- Increase size with interpolation (if needed) so that the shape after convolution will be as-close-as-possible to the output shape
- n_layers of size-increasing convolutional blocks
- Decrease size with interpolation (if needed) so that the shape *exactly* matches the output shape
- Convolve to number of output channels (if needed) so that the number of channels *exactly* matches the output channels

## Results

Running with default settings uses the following layers:

```
CNNEncoder (osisaf-south) with 3 layers
- ResizingInterpolation from (432, 432) to (1024, 1024)
- ConvBlockDownsample (ReLU, 3) with 1 channels
- ConvBlockDownsample (ReLU, 3) with 2 channels
- ConvBlockDownsample (ReLU, 3) with 4 channels
CNNEncoder (era5) with 3 layers
- ResizingInterpolation from (181, 720) to (1024, 1024)
- ConvBlockDownsample (ReLU, 3) with 27 channels
- ConvBlockDownsample (ReLU, 3) with 54 channels
- ConvBlockDownsample (ReLU, 3) with 108 channels
CNNDecoder (osisaf-south) with 3 layers
- ConvBlockUpsample (ReLU, 3) with 224 channels
- ConvBlockUpsample (ReLU, 3) with 112 channels
- ConvBlockUpsample (ReLU, 3) with 56 channels
- ResizingInterpolation from (1024, 1024) to (432, 432)
- Channel convolution from 28 to 1
```

**CNN-UNet-CNN**
cnn-11-100epochs-f5c4547

**CNN-ViT-CNN**
media_images_sea-ice_concentration-static-maps_0_914c2320bbb8666d51c4

Note that there's still a bit of aliasing. I think @louisavz might have some thoughts about how to improve this?

## Results with anti-aliasing

**CNN-UNet-CNN**
media_images_sea-ice_concentration-static-maps_0_4c56074e6e213103f921

**CNN-ViT-CNN**
media_images_sea-ice_concentration-static-maps_0_a0cd9455d817619c384b

### DIFF
--- a/ice_station_zebra/config/model/cnn_vit_cnn.yaml
+++ b/ice_station_zebra/config/model/cnn_vit_cnn.yaml
@@ -1,0 +1,23 @@
+_target_: ice_station_zebra.models.EncodeProcessDecode
+
+name: cnn-vit-cnn
+
+encoder:
+  _target_: ice_station_zebra.models.encoders.CNNEncoder
+  kernel_size: 3 # Size of the kernel for convolutional layers
+  latent_space: [192, 192] # Shape of the latent space
+  n_layers: 3 # Number of convolutional layers
+
+processor:
+  _target_: ice_station_zebra.models.processors.VitProcessor
+  patch_size: 16
+  emb_dim: 128
+  depth: 3
+  heads: 4
+  mlp_dim: 256
+  dropout: 0.3
+
+decoder:
+  _target_: ice_station_zebra.models.decoders.CNNDecoder
+  kernel_size: 3 # Size of the kernel for convolutional layers
+  n_layers: 3 # Number of convolutional layers
--- a/ice_station_zebra/config/model/naive_vit_naive.yaml
+++ b/ice_station_zebra/config/model/naive_vit_naive.yaml
@@ -1,6 +1,6 @@
 _target_: ice_station_zebra.models.EncodeProcessDecode
 
-name: encode-vit-decode
+name: naive-vit-naive
 
 encoder:
   _target_: ice_station_zebra.models.encoders.NaiveLinearEncoder
@@ -8,15 +8,12 @@ encoder:
 
 processor:
   _target_: ice_station_zebra.models.processors.VitProcessor
-  img_size: 192
   patch_size: 16
   emb_dim: 128
   depth: 3
   heads: 4
   mlp_dim: 256
   dropout: 0.3
-  start_out_channels: 28 # Initial number of channels
-
 
 decoder:
   _target_: ice_station_zebra.models.decoders.NaiveLinearDecoder
--- a/ice_station_zebra/models/processors/vit.py
+++ b/ice_station_zebra/models/processors/vit.py
@@ -20,26 +20,29 @@ from ice_station_zebra.types import TensorNCHW
 from .base_processor import BaseProcessor
 
 
-# class VitProcessor(nn.Module):
 class VitProcessor(BaseProcessor):
-    def __init__(  # noqa: PLR0913
+    def __init__(
         self,
-        start_out_channels: int,
-        img_size: int,
-        patch_size: int,
-        emb_dim: int,
+        *,
         depth: int,
+        dropout: float,
+        emb_dim: int,
         heads: int,
         mlp_dim: int,
-        dropout: float,
+        patch_size: int,
         **kwargs: Any,
     ) -> None:
         """Initialize Vision Transformer model for sea ice forecasting."""
         super().__init__(**kwargs)
 
-        self.img_size = img_size
+        # Ensure input is square
+        if self.data_space.shape[0] != self.data_space.shape[1]:
+            msg = "The height and width of the input are not equal."
+            raise ValueError(msg)
+
+        self.img_size = self.data_space.shape[0]
         self.patch_size = patch_size
-        self.out_channels = start_out_channels
+        self.out_channels = self.data_space.channels
 
         self.patch_embed = PatchEmbedding(
             self.data_space.channels, patch_size, emb_dim, self.img_size


louisavz:
Hi @jemrobinson, regarding aliasing and the blockiness of the output [`nn.functional.interpolate`](https://docs.pytorch.org/docs/stable/generated/torch.nn.functional.interpolate.html) has an optional `antialias=False` might worth looking into as an optional input argument (currently not an option in [L25](https://github.com/alan-turing-institute/ice-station-zebra/blob/1542b6c030b030d0bfa5e1e266009a7366c5e251/ice_station_zebra/models/common/resizing_interpolation.py#L25C16-L27C10) of `resizing_interpolation.py`). Another note is I used resize (with Gaussian) for every up convolutional layers. In Tensorflow:

```
tf.image.resize(
            inputs,
            size=[out_height, out_width],
            method=tf.image.ResizeMethod.GAUSSIAN
)
```

ice_station_zebra/models/decoders/cnn_decoder.py
```diff
@@ -36,31 +41,72 @@ def __init__(
...
+            max(minimal_input_shape[0], self.data_space_in.shape[0]),
+            max(minimal_input_shape[1], self.data_space_in.shape[1]),
+        )
+        if shape != self.data_space_in.shape:
+            layers.append(ResizingInterpolation(shape))
+            logger.debug(
+                "- ResizingInterpolation from %s to %s",
+                self.data_space_in.shape,
+                minimal_input_shape,
+            )
```
louisavz, line 75: I'm confused with why this is necessary. There is already a ResizingInterpolation in place at the end, why are we resizing up to minimal_input_shape before starting the conv layers? Why is a minimal_input_shape necessary?

jemrobinson:
I think anti-aliasing has helped a bit here.

ice_station_zebra/models/encoders/cnn_encoder.py
```diff
@@ -33,28 +37,38 @@ def __init__(
...
-
-        # Set the spatial dimensions as required by the number of convolutional layers
+        # If necessary, apply a convolutional resizing to get the correct input dimensions
         initial_required_shape = (
             self.data_space_out.shape[0] * (2**n_layers),
             self.data_space_out.shape[1] * (2**n_layers),
         )
-        layers.append(ResizingInterpolation(initial_required_shape))
+        if self.data_space_in.shape != initial_required_shape:
+            layers.append(ResizingInterpolation(initial_required_shape))
```
louisavz, line 48: I think this is similar to the decoder's minimal_input_shape question above. Why force the initial_required_shape to be a multiple of 2**n_layers. Are you avoiding the use of padding (or other options) for the convolution?

louisavz (commented):
Hi @jemrobinson, I have some questions and rather than me guessing the intents, perhaps you could help me understand better.

louisavz:
> I think anti-aliasing has helped a bit here.

Yea, looks like it, at least, for the background. Not sure what happened with the ViT's. Well done!

louisavz:
Although the range is a bit concerning...

ice_station_zebra/models/decoders/cnn_decoder.py
jemrobinson: The plan was to do the crude size-increasing interpolation before the convolutional layers, so that the "added" information will get smoothed out by the convolution. This means that the final resizing interpolation layer should always be *size-decreasing* which is (hopefully) less likely to produce artifacts. I'm not sure whether this is actually true though - I can try with just the final resizing layer.

ice_station_zebra/models/encoders/cnn_encoder.py
jemrobinson: This is the shape that will give the correct output shape after convolution, since the ConvBlockDownsample layers each reduce (H, W) by 2 while doubling C. I'll see whether doing a simple resize at the end of the process performs well.

ice_station_zebra/models/common/resizing_interpolation.py
```diff
@@ -23,5 +23,9 @@ def __init__(self, output_shape: Sequence[int]) -> None:
 
     def forward(self, x: TensorNCHW) -> TensorNCHW:
         return nn.functional.interpolate(
-            x, size=self.output_shape, mode="bilinear", align_corners=True
+            x,
+            size=self.output_shape,
+            mode="bilinear",
+            align_corners=True,
+            antialias=True,
```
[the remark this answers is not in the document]
jemrobinson: Done in 88017c0

ice_station_zebra/models/decoders/cnn_decoder.py
jemrobinson: Would it make sense to do e.g. a centre-crop instead of a resize for the final size-reducing step?

louisavz: If we do a centre-crop, wouldn't that make the region out of scale? Each pixel has distance attached to it. If our input is say 432x432, then the output should be 432x432. If the last layer output is say 562x562, the information is contained inside those 562x562 pixels representing information for 432x432, and needs to be shaped back accordingly. I'm not sure if cropping makes sense. Shouldn't we be making the final conv layer outputs the expected shape so resize is not necessary?

louisavz:
Regarding range, I need to dive in the code, right now I'm only seeing snippets and trying to take a best guess. Let's keep to what we discussed on slack about sigmoid, as I'm worry that adding sigmoid might mask potential issues and we might forget to dig in deeper. For the ViT, the range looks like they could be floating points range. Do you want to merge this PR as is, and try or add other methods with a new (draft) PR?

louisavz (approved):
LGTM and it has achieved "better than we had before".

PR #118, opened 2025-10-07
Fix SIC downloads

jemrobinson:
SIC downloads were failing because the IceNet downloader looks for specific mask files which are not always available on the FTP server.

This PR workarounds this bug by pre-downloading one of a set of identical files and saving it to the place that the IceNet downloader expects.

### DIFF
--- a/ice_station_zebra/data_processors/preprocessors/icenet_sic.py
+++ b/ice_station_zebra/data_processors/preprocessors/icenet_sic.py
@@ -1,5 +1,7 @@
 import logging
 import os
+from contextlib import suppress
+from ftplib import FTP, error_perm
 from pathlib import Path
 
 import pandas as pd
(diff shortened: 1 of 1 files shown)


louisavz:
HI @jemrobinson, can you explain what do you mean by "one of a set of identical files" and why that would fix the issue? Are you saying to download the mask files separately?

jemrobinson:
All the mask files for a given month are the same. IceNet therefore arbitrarily choses "day = 2" and tries to download that mask. However, day=2 is missing for some months (e.g. Feb 1979) so the downloader fails. However, the file for day=1 does exist and is identical. If we pre-download the day=1 file and name it in the same way as day=2 then the IceNet downloader will skip trying-and-failing to download it from osisaf.

ice_station_zebra/data_processors/preprocessors/icenet_sic.py
```diff
@@ -35,23 +37,65 @@ def download(self, preprocessor_path: Path) -> None:
...
-            south=self.is_south,  # Use mask for the Southern Hemisphere
-            parallel_opens=False,  # Concatenation is not working correctly with dask
-        )
-        sic.download()
+        # Treat each year separately
+        for year in sorted({date.year for date in self.date_range}):
+            logger.info("Preparing to download data for %s.", year)
+            # Generate polar masks: note that only 1979-2015 are available
+            masks = Masks(north=self.is_north, south=self.is_south)
+            mask_year = max(1979, min(year, 2015))
```
louisavz, line 45: Apologies, why 2015?

louisavz: Sorry, just saw your comment on line 43 that only 1979-2015 are available. Please ignore.

PR #120, opened 2025-10-08
Initial fix for patch related checkerboard effect in ViT

erinuclkwon:
Independent predictions for each patch were originally stitched together without considering boundaries between patches, causing a checkerboard effect. An initial attempt using a smaller patch size and adding a convolutional smoothing layer largely removed this effect.

### DIFF
--- a/ice_station_zebra/models/processors/vit.py
+++ b/ice_station_zebra/models/processors/vit.py
@@ -66,6 +66,10 @@ class VitProcessor(BaseProcessor):
             nn.Linear(emb_dim, patch_size * patch_size * self.out_channels),
         )
 
+        self.smooth = nn.Conv2d(
+            self.out_channels, self.out_channels, kernel_size=3, padding=1
+        )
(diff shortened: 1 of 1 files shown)


ice_station_zebra/models/processors/vit.py
```diff
@@ -66,6 +66,10 @@ def __init__(
             nn.Linear(emb_dim, patch_size * patch_size * self.out_channels),
         )
 
+        self.smooth = nn.Conv2d(
```
jemrobinson, line 69: LGTM. Do you want to make the kernel size configurable?

jemrobinson (approved):

ice_station_zebra/models/processors/vit.py
erinuclkwon: Thanks. Yes, I will make it configurable in the next PR after trying out the overlapping suggestion from Louisa.

PR #131, opened 2025-10-20
Fixing not implemented error in zebra train

### LINKED ISSUE #127: Fix not implemented error in zebra train
### Problem / Goal

When I run zebra train on my mac, I get the following error:

> NotImplementedError: The operator ‘aten::_upsample_bilinear2d_aa_backward.grad_input’ is not currently implemented for the MPS device. If you want this op to be considered for addition please comment on https://github.com/pytorch/pytorch/issues/141287 and mention use-case, that resulted in missing op as well as commit hash a1cb3cc05d46d198467bebbb6e8fba50a325d4e7. As a temporary fix, you can set the environment variable `PYTORCH_ENABLE_MPS_FALLBACK=1` to use the CPU as a fallback for this op. WARNING: this will be slower than running natively on MPS.

This seems to have been introduced at the commit c71c5cdca17ec859ac926f5250d31ab274aa6152. 

### What does good look like?

- [ ] Fix code so it runs on a mac

### Related context

_No response_

IFenton:
Close #127

### DIFF
--- a/ice_station_zebra/models/decoders/cnn_decoder.py
+++ b/ice_station_zebra/models/decoders/cnn_decoder.py
@@ -95,7 +95,9 @@ class CNNDecoder(BaseDecoder):
 
         # If necessary, resize downwards to match the output shape
         if shape != self.data_space_out.shape:
-            layers.append(ResizingInterpolation(self.data_space_out.shape, antialias=antialias))
+            layers.append(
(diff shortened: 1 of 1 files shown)


IFenton:
Now tested on Baskerville, and it seems to be running fine

LydiaFrance (approved):
Looks good, I think if someone in the future can figure out how to clean up the configs but it may be a necessary evil. Works on my mac 👍

Issue #103, opened 2025-09-24
Update evaluation plots


jemrobinson:
### Problem / Goal

It would be useful if the evaluation plots included some information about the data used to produce them. For example:

- amount of training data
- number of training epochs

I'm not sure whether this is available at evaluation time (e.g. through the metadata that's saved in the checkpoint) or whether this would only be applicable to plots made at training time.

### What does good look like?

- [ ] Information about training added to plots

PR #137, opened 2025-10-29
Add metadata to Plot Titles

### LINKED ISSUE #123: Add more informative title to evaluation plots
### Problem / Goal

Current plots have a a date on top but it is not evident what that date refers to.
We want to have a better title / text elsewhere on the plot that immediately makes clear what we are looking at
At minimum it should mention what date range the model was trained on, what date the plotted data is showing and what metric is being plotted (SIC)

### What does good look like?

- [ ] the plot has a clear title that mentions the metric & the hemisphere)
- [ ] the plot clearly depicts the timerange for which the data is plotted (eg test date)
- [ ] [optional] the plot also documents what data the model producing those predictions was trained on

### Related context

_No response_

LydiaFrance:
Closes #123 #103 

More complex than I first thought, but managed to create a file in callbacks which looks through the Hydra config files to provide information about training and evaluation within the plot title. Apologies that this is a larger PR than I intended. 

image

### _Caveat..._

I have tried my best here but there is probably a better method for scraping the correct info through Hydra. I am not the best at knowing what might be the most useful info and checking it grabbed the right thing. Given our time limits for the showcase might be good to check this at a later date.

### DIFF
--- a/ice_station_zebra/callbacks/metadata.py
+++ b/ice_station_zebra/callbacks/metadata.py
@@ -28,6 +28,7 @@ class Metadata:
         cadence: Training data cadence string (if available).
         n_points: Number of training points calculated from date range and cadence.
         vars_by_source: Dictionary mapping dataset source names to lists of variable names.
+        n_history_steps: Number of history steps used as model input window (days).
 
     """
 
@@ -37,6 +38,7 @@ class Metadata:
     end: str | None = None
     cadence: str | None = None
     n_points: int | None = None
+    n_history_steps: int | None = None
     vars_by_source: dict[str, list[str]] | None = None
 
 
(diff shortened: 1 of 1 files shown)


LydiaFrance:
Moved metadata to a footer instead. Made the footer optional.

## Video

sea-ice_concentration-video-maps_1_241ea6f8743c2e0bf5be

## Horizontal Colourbars
image (3)

## Vertical Colourbars
image (2)

## Without Footer
image

ice_station_zebra/config/evaluate/callbacks/plotting.yaml
```diff
@@ -23,3 +23,4 @@ plotting:
     vmax: 1.0
     colourbar_location: "horizontal"
     colourbar_strategy: "separate"  # shared | separate (prediction gets its own colourbar)
+    include_footer_metadata: false
```
IFenton: Wondering if we should set the default as True. Given most of the time the plots will be in W&B / local runs where it'd probably be useful to have the info at a glance?

PR #141, opened 2025-11-07
adding sigmoid to decoders

erinuclkwon:
Adding sigmoid to forward() in cnn decoder to constrain predictions to physcially valid sea ice concentration range of [0,1].


IFenton (commented):
I've extended this code, so that now you have the option to apply a sigmoid function (by default it is turned off). I've also added a test to check it works as expected.

At the moment, if you want to bound your outputs between 0 and 1, then you can set the option for the decoder in your local yaml:
```
model: 
  decoder: 
    bounded: true  
```

This works for both the naive and the cnn decoder

PR #152, opened 2025-12-02
93 ddpm add additional blocks

marianovitasari20:


ice_station_zebra/models/ddpm.py
```diff
@@ -0,0 +1,436 @@
...
+            test_metrics[f"test_sieerror_{i}"] = SIEError(leadtimes_to_evaluate=[i])
+        self.test_metrics = MetricCollection(test_metrics)
+
+        self.save_hyperparameters()
+
+    def forward(self, inputs: dict[str, TensorNTCHW]) -> TensorNTCHW:
+        """Generate a single NCHW output with diffusion.
+
+        Args:
+            inputs: TensorNCHW with (batch_size, n_latent_channels_total, latent_height, latent_width)
```
IFenton: Again, not latent space?

ice_station_zebra/models/metrics/sie_error.py
```diff
@@ -0,0 +1,67 @@
...
+        self.pred_sie += (
+            preds[:, :, :, self.leadtimes_to_evaluate].sum().to(self.pred_sie.dtype)
+        )
+        self.true_sie += (
+            target[:, :, :, self.leadtimes_to_evaluate].sum().to(self.true_sie.dtype)
+        )
+
+    def compute(self) -> torch.Tensor:
+        """Compute the final Sea Ice Extent error in km²."""
+        return (self.pred_sie - self.true_sie) * 25**2  # each pixel is 25x25 km
```
IFenton: I think the pixel size should be a variable that can be set rather than hard coded (e.g. AMSR2 works at 6km resolution)

IFenton (commented):
We discussed this in coworking. I've noted few things that it would be good to fix before merging. At the moment Baskerville is down, so I can't test that this runs, but I've done that and these points are addressed it should be fine to merge (with the recognition that this is still a work in progress, and there is more to do in terms of refactoring)

marianovitasari20:
> We discussed this in coworking. I've noted few things that it would be good to fix before merging. At the moment Baskerville is down, so I can't test that this runs, but I've done that and these points are addressed it should be fine to merge (with the recognition that this is still a work in progress, and there is more to do in terms of refactoring)

Thank you so much, @IFenton :). This is very useful. I will adjust based on your feedback as soon as Baskerville is back. As you mentioned, it’s still down.

IFenton:
I've just checked again, and the `evaluate` function now runs :tada:, though it doesn't produce plots. The logger states: `Expected 5D 5D tensors [B,T,C,H,W]; got target=(2, 2, 432, 432), prediction=(2, 2, 432, 432)`. Is that what you expect? Do you know what the first two dimensions are in this final vector?

PR #161, opened 2025-12-15
Delete ice_station_zebra/config/demo.yaml

IFenton:
This was an early version of the yaml needed to run the demo notebook. It was replaced by demo_nb.yaml, so it should be deleted


aranas (approved):
makes sense!

PR #167, opened 2026-01-14
[WIP] Shifting to IceNet v0.4.0

### LINKED ISSUE #145: Update sea ice preprocessing to use IceNet v0.4.0
### Problem / Goal

At the moment we are using [IceNet v0.2.9](https://github.com/icenet-ai/icenet) to handle the sea ice downloading / preprocessing. There is a later version of [IceNet v0.4.0](https://github.com/icenet-ai/icenet/tree/v0.4.0_dev), so we should shift to that. (N.b. v0.4.0 is still a branch so is not accessible through PyPI)

In v0.4.0 some of the code is actually under the [environmental forecasting download toolbox](https://github.com/environmental-forecasting/download-toolbox), so it might be that we actually need that rather than the core IceNet library

### What does good look like?

- [ ] Update ISZ code to use IceNet v0.4.0

### Related context

_No response_

IFenton:
Want to update our code so that it runs IceNet v0.4.0 rather than the current 0.2.7

Closes #145


IFenton:
Closing this, as the work in #203 removes the dependence on the IceNet tooling for accessing the OSISAF data

PR #217, opened 2026-02-13
Update ERA5 config files

### LINKED ISSUE #202: Reduce amount of ERA5 data
### Problem / Goal

We're currently downloading the full hemisphere of ERA5 data but the EASE2 grid is 432x432 25km cells. This corresponds to a maximum distance of `(432 / 2) * 25km * sqrt(2) ~- 7640km` from the pole. This corresponds to about 21 degrees of latitude, so we could consider reducing our ERA5 data to 21 degrees+ which should reduce the necessary data by ~20%.

### What does good look like?

- [x] Decision about how much ERA5 data to download.

### Related context

_No response_

jemrobinson:
I've added some new ERA5 configs which I've set as default (but not removed the old configs yet). These do the following:

- Reduce latitudinal extent to 20-90 degrees
- :bug: Changed frequency from `1d` to `24h` as this was causing Anemoi warnings
- Switched to finest available grid (0.25x0.25 degrees i.e. 4x the amount of data we had before)
- Added `insolation` as a potentially-useful variable

Here's a recent W&B run using this config file: https://wandb.ai/turing-seaice/naive-unet-naive/runs/teba39a7?nw=nwuserjamesrobinson

Closes #202

### DIFF
--- a/icenet_mp/config/data/datasets/samp_weathernorth_era5_0p5_2017_2019_24h_v2.yaml
+++ b/icenet_mp/config/data/datasets/samp_weathernorth_era5_0p5_2017_2019_24h_v2.yaml
@@ -1,7 +1,7 @@
-samp-weathersouth-era5-0p25-2017-2019-24h-v1:
-  name: samp-weathersouth-era5-0p25-2017-2019-24h-v1
+samp-weathernorth-era5-0p5-2017-2019-24h-v2:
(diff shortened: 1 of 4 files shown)


jemrobinson:
Merging based on approval from @marianovitasari20 and @IFenton

PR #222, opened 2026-02-19
Fix FFMPEG issue

### LINKED ISSUE #215: Remove ffmpeg dependency
### Problem / Goal

Recent plotting changes made in #146 added a dependency on ffmpeg to make mp4 video instead of the previous gif format. This is not installed on some HPC systems (e.g. IsambardAI) which causes the error:

```
Video plotting skipped: Video encoding failed: [Errno 2] No such file or directory: 'ffmpeg'
Video plotting skipped: Video encoding failed: [Errno 2] No such file or directory: 'ffmpeg'
Video plotting skipped: Video encoding failed: [Errno 2] No such file or directory: 'ffmpeg'
```

### What does good look like?

- [ ] Revert ffmpeg dependency

### Related context

- Caused by #146

jemrobinson:
- use `imageio-ffmpeg` to install FFMPEG
- set the appropriate `matplotlib` configuration to allow this to be used by default

Closes #215

P.S. I haven't tested this - if someone who's experiencing the problem can do so, that would be great.


jemrobinson:
I think the best way to test is to do a minimal `train` run (e.g. with a simple model and 1 or 2 epochs) and then run `evaluate` on the checkpoint from that run. Evaluate is the step that runs the video generation.

- It should not print the line `Video plotting skipped: Video encoding failed: [Errno 2] No such file or directory: 'ffmpeg'`. 
- It should produce the `.mp4` video which gets uploaded to W&B.

jemrobinson:
@IFenton if you have time but I'd rather have your focus on getting #216 merged. This PR fixes (or attempts to fix) something that's currently broken but doesn't seem to be problematic.

PR #241, opened 2026-03-09
Standardise configs/models for comparison

jemrobinson:
- Standardised parameters across:
  - cnn_null_cnn
  - cnn_unet_cnn
  - cnn_vit_cnn
  - ddpm
  - naive_null_naive
  - naive_unet_naive
  - naive_vit_naive
  - persistence
  - piecewise_null_piecewise
  - piecewise_unet_piecewise
  - piecewise_vit_piecewise
- Added 'hemisphere' as a config option
  - This allows us to filter on it for train/evaluate jobs
- Added type hints
- Added configs for 14 day and 21 day predictions
- Use base `test_metrics`, `test_loss` and `val_loss` in DDPM for easier comparison with other models


IFenton (approved):
With your changes to the default latent space size etc. this now runs very slowly on my computer (0.11 it/s vs. 3.74 it/s). Would make sense to change some of those defaults for the sample runs?

I'm also curious as to why you changed how the hemisphere was being obtained?

IFenton (approved):
LGTM. Would be good to add a quick note to the README about the new `quick_test` config

Issue #247, opened 2026-03-30
Evaluate impact of adding Argo float data


jemrobinson:
### Problem / Goal

How much does the Argo float data help? What accuracy/precision etc. did we have before? What do we have afterwards?

### What does good look like?

- [ ] Can quantify impact of adding Argo float data

IFenton:
As part of this, we should test our assumptions about the time window (4hrs) / water depth (0-50m) that we are using - are they the optimal values?

IFenton:
Some more thoughts on testing this: 
* We’ve already discussed better ways of gridding data - the method we use is very crude
* To do a better comparison with our models we should be including a gridded version of sea surface temperature / salinity in the data as otherwise we’re not really comparing like with like
* I also suspect we should check again after we’ve investigated the grid reprojections / masking. Have we actually checked that adding ERA5 makes improvements to our model in the current setup? It should, but my current impression is that everything is dominated by the SIC at the moment.

aranas:
To evaluate if change in performance is meaningful, increase the density of argo float data incrementally and see if the reduction in error behaves as expected (eg error roughly decreases with increasing amount of obs data).

PR #248, opened 2026-04-07
Minor fixes related to Argo float running

jemrobinson:
# Question
Should we add Argo float data to the default setup?

# Testing method
- run over sample datasets with or without inclusion of sample Argo data
- simple model: naive-unet-naive with n_parameters:
  - encoder_sic_icenet: 2
  - encoder_era5: 56
  - encoder_float_argo: 4
  - processor: 11.0M
  - decoder: 32
- local run on MPS
- 50 epochs

# Results
| | No Argo (baseline) | With Argo | Persistence |
| -- | -- | -- | -- |
| SIE error | **481134** | 1370787 | _293761_ |
| RMSE | **0.07688** | 0.08967 | _0.02352_ | 
| MAE | 0.02199 | **0.01893** | _0.00320_ | 
| Train it/s | 3.82 | **3.98** | _35.35_ |

sie_error
rmse
mae

## No Argo
no-argo

## With Argo
with-argo

## Persistence
persistence

### DIFF
--- a/icenet_mp/config/data/datasets/samp_floatsouth_argo_pt_2017_2019_24h_v1.yaml
+++ b/icenet_mp/config/data/datasets/samp_floatsouth_argo_pt_2017_2019_24h_v1.yaml
@@ -1,19 +1,19 @@
-samp-floatnorth-argo-pt-2017-2019-24h-v1:
-  name: samp-floatnorth-argo-pt-2017-2019-24h-v1
+samp-floatsouth-argo-pt-2017-2019-24h-v1:
+  name: samp-floatsouth-argo-pt-2017-2019-24h-v1
   group_as: float_argo
-  description: ARGO float data for northern hemisphere
+  description: ARGO float data for southern hemisphere
   attribution: ARGO
   licence: CC-BY-4.0
 
   dates:
     start: '2017-01-01T12:00:00'
-    end: '2017-01-20T12:00:00'
+    end: '2019-01-31T23:00:00'
     frequency: 24h
 
   input:
     pipe:
       - argo:
-          area: "90/-180/20/180"  # northern hemisphere
+          area: "-20/-180/-90/180"  # southern hemisphere
           param:
             - TEMP # temperature
             - PSAL # salinity
--- a/icenet_mp/config/data/datasets/samp_floatsouth_argo_pt_2017_2019_24h_v1.yaml
+++ b/icenet_mp/config/data/datasets/samp_floatsouth_argo_pt_2017_2019_24h_v1.yaml
@@ -7,7 +7,7 @@ samp-floatsouth-argo-pt-2017-2019-24h-v1:
 
   dates:
     start: '2017-01-01T12:00:00'
-    end: '2019-01-31T23:00:00'
+    end: '2019-01-31T12:00:00'
     frequency: 24h
 
   input:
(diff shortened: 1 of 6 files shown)


IFenton:
Reviewing this

icenet_mp/config/data/datasets/samp_floatsouth_argo_pt_2017_2019_24h_v1.yaml
```diff
@@ -1,19 +1,19 @@
...
+samp-floatsouth-argo-pt-2017-2019-24h-v1:
+  name: samp-floatsouth-argo-pt-2017-2019-24h-v1
   group_as: float_argo
-  description: ARGO float data for northern hemisphere
+  description: ARGO float data for southern hemisphere
   attribution: ARGO
   licence: CC-BY-4.0
 
   dates:
     start: '2017-01-01T12:00:00'
```
IFenton, line 9:
```suggestion
    start: '2017-01-01T00:00:00'
```

IFenton (approved):
LGTM (though I don't think it closes #247)

icenet_mp/config/data/datasets/samp_floatsouth_argo_pt_2017_2019_24h_v1.yaml
jemrobinson: I don't think we want this. We're taking data from close-to 12:00, right?

jemrobinson:
> LGTM (though I don't think it closes https://github.com/alan-turing-institute/icenet-mp/issues/247)

What more would you like to see before we could close #247?

icenet_mp/config/data/datasets/samp_floatsouth_argo_pt_2017_2019_24h_v1.yaml
IFenton: I was updating it to match the other datasets given you had updated the end time to 23:59. But I don't think it makes any difference to the data that is downloaded. We're currently taking the data from 10:00-14:00.

IFenton:
> What more would you like to see before we could close https://github.com/alan-turing-institute/icenet-mp/issues/247?

Sophie and I added some comments to that issue about what it would be good to do to test the impact of adding the Argo float data (e.g. testing assumptions about depth / time window, gridding mechanisms etc.). I don't think they need to be done as part of this PR, but if we're going to close #247 we should capture those comments in a new issue about improving the Argo float data. And I would also update the title of #247 to reflect this (e.g. Evaluate impact of current implementation for adding Argo float data)

icenet_mp/config/data/datasets/samp_floatsouth_argo_pt_2017_2019_24h_v1.yaml
```diff
@@ -1,19 +1,19 @@
...
-    end: '2017-01-20T12:00:00'
+    end: '2019-01-31T23:00:00'
```
jemrobinson:
```suggestion
    end: '2019-01-31T12:00:00'
```

icenet_mp/config/data/datasets/samp_floatsouth_argo_pt_2017_2019_24h_v1.yaml
jemrobinson: Ah OK. I've changed the end datetime to match instead as I think this is more correct.

jemrobinson:
Merging as set of fixes to Argo source but nothing related to changing defaults.

PR #251, opened 2026-04-16
Active grid cell mask in decoder

IFenton:
I've been working on integrating the active grid cell mask into the decoder (rather than in the loss function). It looks like it improves the results, though I also found that with the short training runs I'm doing there is a lot of variation between runs making hard to be certain what the improvements are from a single run comparison. 

Screenshot 2026-04-16 at 12 08 00

AG44 - active grid cell mask 4 days history 4 days forecast 4 epochs
None 44 - no mask 4 days...


jemrobinson:
I was thinking that we ought to do this in a way that can be easily applied to all decoders. This could be done for "bounded" at the same time, so there's (for example) something like this:

```python
class BaseDecoder:
    ...
    def finalise(self, x: TensorNCHW) -> TensorNCHW:
        x = apply_mask(x)
        x = restrict_to_range(x)
        return x
```

```python
class MyDecoderChildClass:
    ...
    def forward(self, x: TensorNCHW) -> TensorNCHW:        
        x = self.model(x)
        x = self.finalise(x)  # inherited from base class
        return x
```

Shaerdan:
hi @IFenton           I've pushed a commit  ( ([d0ae392](https://github.com/alan-turing-institute/icenet-mp/pull/251/commits/d0ae39224993d8435330dba582f32a8c93b438c1)) ) on top of your branch that does the integration/refactor side of this. Here is a summary:
_What was covered:_ 
1.   derives the mask path from the SIC dataset (wires in our in-house masks #257/#266, drops the broken hardcoded path); 
2.   unifies apply into one shared BaseDecoder.finalise() (ref James's note); 
3.   fixes the ordering so masked cells are exactly 0 (mask after sigmoid, avoiding sigmoid(0)=0.5), conditioned on use_mask flag.

_I tested it by these:_
1. the 6 CI tests passed clean
2. real-mask integration test: against the actual 432×432 SSMIS mask, the decoder clamped inactive-masked cells to exact 0, across multiple random inputs.
3. back compatibility: use_mask=false the output is byte-identical to main.

some questions left for discussion: the sic-ssmis encoder tag (config-only, ties into #9/#206) is not set yet in the encoders' config files, do we plan to move to ssmis and away from OSISAF? if so i can update the default config for all encoders to use sic-ssmis instead.

Could you take a look and take it forward (mark ready / request James) if it looks right?

IFenton:
@Shaerdan Have you tested a full model run with this integrated? I would be interested in looking at the WandB plots.

Shaerdan:
@IFenton not yet, i am very tempted to give that a try. I might submit a job later and run it for 10/20 epochs, also need to figure out how to get the wandb plots here, and then i'll ping you tomorrow if the results are ready

IFenton:
I was having issues testing this with SSMIS data as it was giving an error, but that bug is now fixed #331

IFenton:
I've tested the active grid cell mask / the bounded outputs:
Mask and bounded: https://wandb.ai/turing-seaice/evaluate/runs/51vzl0ro
Mask and not bounded: https://wandb.ai/turing-seaice/evaluate/runs/hr5r29oq
No mask and bounded: https://wandb.ai/turing-seaice/evaluate/runs/2ov1sqx5
No mask, not bounded: https://wandb.ai/turing-seaice/evaluate/runs/qrrpd9fb

icenet_mp/models/decoders/base_decoder.py
```diff
@@ -40,6 +68,19 @@ def forward(self, x: TensorNCHW) -> TensorNCHW:
...
         raise NotImplementedError(msg)
 
+    def finalise(self, x: TensorNCHW) -> TensorNCHW:
+        """Apply shared output steps: bound if requested, then zero masked cells.
+
+        Masking is applied AFTER bounding so masked cells are exactly 0 regardless of
+        `bounded` (applying it before would leak sigmoid(0)=0.5 into masked cells).
+        Every concrete decoder should call this at the end of its `forward`.
+        """
+        if self.bounded:
```
jemrobinson: Could we use `RangeRestriction` as done in the `PiecewiseDecoder` here? Happy for it to default to sigmoid though.

tests/models/test_decoders.py
```diff
@@ -210,3 +313,43 @@ def test_clamp_restricts_output_to_unit_range(self) -> None:
...
         assert torch.all(output <= 1.0).item()
+
+
+@pytest.mark.skipif(
+    not _REAL_MASKS, reason="No real SSMIS active mask available locally"
+)
+class TestDecoderMaskOnRealMask:
+    """Integration check against the real 432x432 active grid cell mask on disk."""
+
+    def test_decoder_zeros_exactly_the_inactive_cells(self) -> None:
```
jemrobinson, line 347: This can actually fail if you happen to generate exactly 0 with torch.randn in the same place across each of your random samples. I think this is unlikely enough not to worry about though.

icenet_mp/data_loaders/common_data_module.py
```diff
@@ -119,6 +120,18 @@ def longitudes(self) -> dict[str, list[float]]:
...
+    def active_mask_path(self) -> Path:
+        """Path to the active mask for the dataset.
+
+        Derived from the `base_path` (root) and the SIC target dataset (mask filename),
+        so the path always traces the data path. Note: the file exists only for
+        datasets whose masks have been generated (eg, currently the SSMIS datasets
+        sitting on the isambard shard workspace).
+        """
+        target_name = self.dataset_groups[self.target_group_name][0].stem
+        return mask_dir(self.base_path, target_name) / "active_mask.npy"
```
jemrobinson: When we write to the `mask_dir` we use key `name` where `name` is the outer config key (the dict key under `config["data"]["datasets"]`) but when we're reading it here we use `stem` which comes from the inner `dataset['name']` field. We should be consistent.

icenet_mp/data_loaders/common_data_module.py
```diff
@@ -119,6 +120,18 @@ def longitudes(self) -> dict[str, list[float]]:
...
+    @cached_property
+    def active_mask_path(self) -> Path:
+        """Path to the active mask for the dataset.
+
+        Derived from the `base_path` (root) and the SIC target dataset (mask filename),
+        so the path always traces the data path. Note: the file exists only for
+        datasets whose masks have been generated (eg, currently the SSMIS datasets
+        sitting on the isambard shard workspace).
+        """
+        target_name = self.dataset_groups[self.target_group_name][0].stem
```
jemrobinson: If we have multiple datasets from the same group, this will always look for the mask from the first one rather than checking to see which are available and combining them if appropriate.

icenet_mp/models/decoders/base_decoder.py
```diff
@@ -27,6 +37,24 @@ def __init__(
...
+        self.use_mask = use_mask
+        if use_mask:
+            if active_mask_path is None or not Path(active_mask_path).exists():
+                msg = (
+                    f"use_mask is enabled but no active mask was found at "
+                    f"{active_mask_path}. Masks are generated per dataset during "
+                    f"`datasets create` (currently for SSMIS datasets)."
+                )
+                raise FileNotFoundError(msg)
+            mask = from_numpy(np.load(Path(active_mask_path))).float()
```
jemrobinson: We should check that mask shape matches `data_space_out.shape` here.

icenet_mp/model_service.py
```diff
@@ -131,6 +132,7 @@ def from_checkpoint(
                 checkpoint_path,
                 latitudes_fn=lambda: builder.data_module.latitudes,
                 longitudes_fn=lambda: builder.data_module.longitudes,
+                active_mask_path=str(builder.data_module.active_mask_path),
```
jemrobinson: This will always overwrite a path that's saved in the checkpoint. That's probably OK since we want this to be portable, but something to be aware of.

icenet_mp/models/decoders/piecewise_decoder.py
```diff
@@ -124,4 +124,5 @@ def forward(self, x: TensorNCHW) -> TensorNCHW:
             TensorNCHW with (batch_size, output_channels, output_height, output_width)
 
         """
-        return self.model(x)
+        output = self.model(x)
+        return self.finalise(output)
```
jemrobinson: Note that this will apply bounding twice (once in the PiecewiseDecoder model and once in finalise). Remove the PiecewiseDecoder bounding (and ideally move it to finalise as it supports multiple bounding types).

icenet_mp/models/decoders/base_decoder.py
```diff
@@ -40,6 +68,19 @@ def forward(self, x: TensorNCHW) -> TensorNCHW:
...
+        `bounded` (applying it before would leak sigmoid(0)=0.5 into masked cells).
+        Every concrete decoder should call this at the end of its `forward`.
+        """
+        if self.bounded:
+            x = sigmoid(x)
+        if self.use_mask:
+            x = x * self.mask.to(dtype=x.dtype)
+        return x
+
     def rollout(self, x: TensorNTCHW) -> TensorNTCHW:
```
jemrobinson, line 69: Should we put finalise() here? Instead of relying on each model's `forward()` to call it?

jemrobinson (changes requested):
Results look encouraging: some thoughts here.

icenet_mp/models/decoders/base_decoder.py
Shaerdan: @jemrobinson good idea, just to confirm to be sure: do you mean, to have finalise() both in each model's forward() and also in finalise(), just in case some models don't have a forward() or have an empty one?  the sub-question is: do all models use roll_out() after their forward()? if they do, then I can just put finalise() in roll_out() and not in forward(); if some don't, then I will put finalise() in both place and put an if condition before finalise() is called in roll_out(). what do you think?

jemrobinson: 1. All decoders need to be able to implement `rollout(self, x: TensorNTCHW) -> TensorNTCHW`, but at the moment they all defer this to the base class implementation in `BaseDecoder`.
2. We want to make sure that `finalise()` is called by all decoders that are trying to predict SIC (but not by those that are trying to predict other variables where this masking or bounding behaviour is unnecessary

I think the best way to achieve this is probably to have two flags on `BaseDecoder` (similar to how `bounded` is used on the `CNNDecoder` now) which control whether to apply bounding and whether to apply masking. This can either be done in the `forward()` method of each derived `Decoder` or in the `BaseDecoder`.

If we do it in the `BaseDecoder` I think it would make sense to do it inside `rollout` which is the function that then calls `forward` on the derived classes. However, allowing each derived class to handle `bounded` and `masked` independently is an option if you think the flexibility is more important than the code duplication.

icenet_mp/model_service.py
Shaerdan: Exactly, this is intentional. We rederive the mask path from base_path so a checkpoint can be portable across data roots or machines.

tests/models/test_decoders.py
Shaerdan: Fixed (partially and brute force), the test now draws until the always-zero set converges to the inactive set of the mask, and only fails if it never converges within 50 draws. I also added a direct test using the mask from the file as a deterministic double check just in case. We probably only need one of the two but doesnt hurt to have both to be thorough.

icenet_mp/models/decoders/base_decoder.py
Shaerdan: resolved

icenet_mp/models/decoders/base_decoder.py
Shaerdan: resolved

icenet_mp/data_loaders/common_data_module.py
Shaerdan: resolved

icenet_mp/data_loaders/common_data_module.py
Shaerdan: resolved

icenet_mp/models/decoders/base_decoder.py
```diff
@@ -57,6 +113,5 @@ def rollout(self, x: TensorNTCHW) -> TensorNTCHW:
             TensorNTCHW with (batch_size, n_forecast_steps, output_channels, output_height, output_width)
 
         """
-        return self(x.reshape(-1, *self.data_space_in.chw)).reshape(
-            -1, self.n_forecast_steps, *self.data_space_out.chw
-        )
+        output = self.finalise(self(x.reshape(-1, *self.data_space_in.chw)))
+        return output.reshape(-1, self.n_forecast_steps, *self.data_space_out.chw)
```
jemrobinson: Should this be:

```
output = self(x.reshape(-1, *self.data_space_in.chw)).reshape(-1, self.n_forecast_steps, *self.data_space_out.chw)
return self.finalise(output)
```

or is there a reason why `finalise` doesn't apply to the reshaped output?

icenet_mp/models/decoders/piecewise_decoder.py
```diff
@@ -38,7 +37,7 @@ def __init__(
         **kwargs: Any,
     ) -> None:
         """Initialise a PiecewiseDecoder."""
-        super().__init__(**kwargs)
+        super().__init__(restrict_range=restrict_range, **kwargs)
```
jemrobinson:
```suggestion
**kwargs: Any,
) -> None:
"""Initialise a PiecewiseDecoder."""
        super().__init__(**kwargs)
```

I don't think we need a decoder-specific default here just for `PiecewiseDecoder`, so let's keep this simple and remove the `restrict_range: str = "clamp",` argument above too.

icenet_mp/models/decoders/base_decoder.py
```diff
@@ -27,6 +38,35 @@ def __init__(
...
+            if mask_path is None or not Path(mask_path).exists():
+                msg = (
+                    f"{mask_type} mask is requested but no mask was found at "
+                    f"{mask_path}. Masks are generated per dataset during "
+                    f"`datasets create` (currently for SSMIS datasets)."
+                )
+                raise FileNotFoundError(msg)
+            mask = from_numpy(np.load(Path(mask_path))).float()
+            if tuple(mask.shape) != self.data_space_out.shape:
+                msg = f"{mask_type} mask shape does not match decoder output"
```
jemrobinson: Stated what the actual and expected shapes are here, to help in debugging.

icenet_mp/callbacks/plotting_callback.py
```diff
@@ -123,6 +122,11 @@ def make_plots(
             return
         self.plotter.set_hemisphere(pl_module.hemisphere)
 
+        # Load land mask for plotting based on dataset
+        datamodule = getattr(trainer, "datamodule", None)
+        land_mask_path = getattr(datamodule, "land_mask_path", None)
+        self.plotter.land_mask = LandMask(land_mask_path)
```
jemrobinson: I think this means that we are re-building the LandMask each time this function is called. Let's try to cache this (probably fine to do so per-instance as we're likely to only get one datamodule, but also fine to do at the per-datamodule level if that feels safer).

jemrobinson (changes requested):
Looks like this is in good shape, but I have a couple of comments in addition to the CoPilot ones.

icenet_mp/models/decoders/base_decoder.py
Shaerdan: Done in 3717637d — the error now includes both shapes

icenet_mp/callbacks/plotting_callback.py
Shaerdan: Done in 3717637d: LandMask is now built once and cached per land_mask_path on the callback instance -- self._land_mask_cache -- so it doesnt rebuilt every validation epoch. Kept per instance.

jemrobinson (changes requested):
This is looking pretty good. A couple of general compatibility fixes needed:

1. We need the "encoderstage" of multistage training to opt out of masking (since we want to learn a full encoding of all input data)
2. We want the standalone DDPM model to opt *into* masking (possible just on the inference pathway).

icenet_mp/models/decoders/base_decoder.py
```diff
@@ -13,18 +17,57 @@ class BaseDecoder(nn.Module):
...
+            RangeRestriction(restrict_range), min_val=0, max_val=1
+        )
+
+        # Load the active mask only when requested. When off, finalise() skips
+        # the multiply entirely. Path is derived from the dataset (ref CommonDataModule.active_mask_path)
+        # Require the file to exist when mask_type is defined, fail loudly if not.
+        if mask_type not in (None, "none", "active", "land"):
+            msg = f"Unknown mask_type {mask_type!r}; expected one of none/active/land."
+            raise ValueError(msg)
+        self.mask_type = mask_type
```
jemrobinson: Do we need this? Do we use `self.mask_type` anywhere?

icenet_mp/models/decoders/base_decoder.py
```diff
@@ -13,18 +17,57 @@ class BaseDecoder(nn.Module):
...
 
+        # Bound (or not) the output into [0, 1], select: none/sigmoid/clamp/tanh.
+        self.restrict = RestrictRange(
+            RangeRestriction(restrict_range), min_val=0, max_val=1
+        )
+
+        # Load the active mask only when requested. When off, finalise() skips
+        # the multiply entirely. Path is derived from the dataset (ref CommonDataModule.active_mask_path)
+        # Require the file to exist when mask_type is defined, fail loudly if not.
+        if mask_type not in (None, "none", "active", "land"):
```
jemrobinson: What's the difference between `None` and `"none"`? Do we need both?

icenet_mp/models/ddpm.py
```diff
@@ -74,24 +81,65 @@ def __init__(  # noqa: PLR0913
...
+                )
+                raise FileNotFoundError(msg)
+            mask = torch.from_numpy(np.load(Path(mask_path))).float()
+            if tuple(mask.shape) != self.output_space.shape:
+                msg = (
+                    f"{mask_type} mask shape {tuple(mask.shape)} does not match "
+                    f"output shape {self.output_space.shape}."
+                )
+                raise ValueError(msg)
+            self.register_buffer("mask", mask, persistent=False)
```
jemrobinson: This is all duplicated from BaseDecoder. How about instead making a Mask layer in models/common that both BaseDecoder and DDPM could use?

icenet_mp/models/decoders/base_decoder.py
IFenton: This is how you choose the type of mask you want to apply (active or land mask) or to not apply a mask.

icenet_mp/models/ddpm.py
IFenton: Sounds sensible

icenet_mp/models/decoders/base_decoder.py
IFenton: I assume this is because the config takes in text arguments, rather than accepting `None` - do you know if there is a better way of doing it?

icenet_mp/models/decoders/base_decoder.py
jemrobinson: I mean, why are we saving the argument `mask_type` as an attribute of the object? I don't think it is used elsewhere.

icenet_mp/models/decoders/base_decoder.py
jemrobinson: You can write a YAML config option like `my_option: null` which will be translated in Python to `my_option = None`.

icenet_mp/models/decoders/base_decoder.py
IFenton: No it's not, agreed it doesn't need to be saved

icenet_mp/models/ddpm.py
jemrobinson: Done in 270dca4

icenet_mp/models/decoders/base_decoder.py
jemrobinson: Done in 270dca4

icenet_mp/models/decoders/base_decoder.py
jemrobinson: Done in 672f3d67

jemrobinson (approved):
Applied some small refactors to simplify the structure. LGTM.

Issue #250, opened 2026-04-14
Add a latent space diagnostic notebook


aranas:
### Problem / Goal

To aid model development we want visual diagnostics of intermediate layer activation in the sea ice prediction pipeline. 

As a first target, we want to visualize the multimodal argo-float pipeline to provide evidence for the following research question:

> Does incorporating Argo float observations shift the model's internal representation in a physically meaningful way and if so, at which layer does that shift first become detectable?

Initial demonstration of this diagnostic will be a standalone diagnostic notebook on a separate branch, integration into pipeline might follow at a later stage if diagnostic proves valuable.

### What does good look like?

- [x] A `latent_space_viz` feature branch on the pipeline repo that adds an optional `--save-activations` flag to the inference CLI, which when set triggers hook registration before the forward pass and saves captured activations to disk as xarray datasets or tensors keyed by layer name and timestep.
- [x] A `viz/hook_manager.py` module that registers forward_hooks on named encoder/processor/decoder modules at runtime, captures activations as the forward pass executes, and writes them to the configured output path.
- [x] A `notebooks/layer_diagnostics.ipynb` notebook that orchestrates the full diagnostic by invoking the pipeline inference CLI via shell cells, once with argo float data, once without, both with `--save-activations` set  and loads the resulting activation files for visualisation.

**Not in scope**
- No changes to model code, configs, or data modules
- No production integration, CLI tooling
- No geographic projection or statistical analysis logic

### Related context

#98

aranas:
@IFenton @jemrobinson 
1. Should this be best done for UNet or viT?
2. which layers to hook? once after encoder, processor (3 layers throughout model, or attention maps for viT), just before & after decoder?
2. Does the diagnostic approach make sense given the research question. Is a with/without Argo activation comparison a reasonable way to probe float data influence?
3. Is attaching external `forward_hook`s to named modules feasible with the current architecture, or are there known gotchas?
4. What documentation or interface conventions would you need from the notebook to make future pipeline integration straightforward?

aranas:
Regarding which model to use:
- I have contacted Erin for a calibrated model checkpoint. If I get this, we could use this to see if diagnostics reflect something meaningful in a reasonable model
- for the argo float data we do not have a fully trained model yet, but we can still inspect encoder changes

aranas:
Discussion with @louisavz @jemrobinson @IFenton 

- concern is that argo float data is not good enough to showcase benefit at this point
[...]

PR #252, opened 2026-04-17
Add optional layer diagnostics features

aranas:
re #250
adding optional feature to save layer activations to disk + notebook to visualize them

- add hook_manager.py file that hooks on selected EncodeProcessDecode submodules and saves activations to disk
- ad hook features as optional via save-activations flag
- add layer_diagnostics notebook to visualize layers


jemrobinson:
👀 looking at this.

jemrobinson:
@aranas : I've made the following changes:

- moved the callback to the callbacks directory (and renamed it to ActivationSaver)
- changed its initialisation so that it reads the model `on_test_start` rather than at initialisation time
- simplified the CLI options so you just need to provide `--save-layer <LAYER_NAME>` one or more time to enable this
- run the callback by default but only do work if there is at least one layer enabled

It might be nice to also save some of these plots to W&B, what do you think?

aranas:
thanks @jemrobinson this works for me. I agree it could be nice to have the plots in wandb but this might be out of scope for #250 and something for the wider team to decide if they want this. I suggest we merge this PR now if you have no other doubts?

jemrobinson:
Yes, W&B integration can come later. Can you update this with the latest main and check that things still work for you?

aranas:
I have updated and notebook still runs fine. notebook has been updated to reflect single flag and default output path

jemrobinson (approved):
LGTM

PR #255, opened 2026-04-20
Add seed

marianovitasari20:
# **Description**

This PR introduces a seed parameter to ensure reproducibility across experiments. By setting a fixed seed, the results become deterministic and easier to debug, compare, and validate.

### **Changes**

* Added global seed
* Ensured consistent behavior across runs

### **Why this is needed**

Reproducibility is important for:

* Debugging model behavior
* Comparing experimental results
* Ensuring consistency across different environments

### DIFF
--- a/icenet_mp/cli/train.py
+++ b/icenet_mp/cli/train.py
@@ -1,7 +1,6 @@
 import logging
 
 import typer
-from lightning import seed_everything
 from omegaconf import DictConfig
 
 from icenet_mp.model_service import ModelService
@@ -18,9 +17,6 @@ log = logging.getLogger(__name__)
 @hydra_adaptor
 def train(config: DictConfig) -> None:
     """Train a model."""
-    if seed := config.get("seed", None):
-        seed_everything(seed, workers=True)
-
     model = ModelService.from_config(config)
     model.train()
 
--- a/icenet_mp/cli/train.py
+++ b/icenet_mp/cli/train.py
@@ -6,6 +6,7 @@ from omegaconf import DictConfig
 from icenet_mp.model_service import ModelService
 
 from .hydra import hydra_adaptor
+from lightning import seed_everything
 
 # Create the typer app
 training_cli = typer.Typer(help="Train models")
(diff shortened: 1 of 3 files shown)


icenet_mp/cli/train.py
```diff
@@ -17,6 +18,9 @@
 @hydra_adaptor
 def train(config: DictConfig) -> None:
     """Train a model."""
+    if seed := config.get("seed", None):
+        seed_everything(seed, workers=True)
```
jemrobinson: I think it might make more sense to do this in the `ModelService` constructor since (a) this file is already importing `lightning` and (b) it should then apply to both `train` and `evaluate`.

icenet_mp/config/base.yaml
```diff
@@ -10,3 +10,9 @@ defaults:
   - _self_
 
 base_path: ../base
+
+seed: 555
+
+train:
+  trainer:
+    deterministic: true
```
jemrobinson: Should this variable be set (a) all the time or (b) only if the seed is set?

If (a) then this should live in `train/trainer/default.yaml`. If (b) then set it in the ModelService `trainer` initialisation, depending on the contents of `seed`.

jemrobinson (changes requested):
Looks good, but I think some of these things should be moved to other places. What do you think?

icenet_mp/cli/train.py
marianovitasari20: thanks @jemrobinson , fair enough, I will change this to `ModelService`

icenet_mp/config/base.yaml
marianovitasari20: This only makes sense if the seed is set. Yes, we can just define the seed in the YAML and have `ModelService` automatically enable deterministic mode when the seed is not None. This should be cleaner. I will change that.

icenet_mp/model_service.py
```diff
@@ -27,6 +27,8 @@ class ModelService:
     def __init__(self, config: DictConfig) -> None:
         """Initialize the model service."""
         self.config_ = config
+        if seed := config.get("seed", None):
+            seed_everything(seed, workers=True)
```
jemrobinson:
```suggestion
            seed_everything(int(seed), workers=True)
```

jemrobinson: Not sure what happens if we give `seed_everything` a string, but probably better not to!

jemrobinson (approved):
LGTM. One small suggestion.

marianovitasari20:
> LGTM. One small suggestion.

Thanks @jemrobinson!

Issue #258, opened 2026-04-23
Investigate validation loss increasing


marianovitasari20:
### Problem / Goal

The validation loss is consistently increasing when using the current default `base` / `sample_dataset` configuration. This behavior needs to be investigated.

### What does good look like?

- [ ] Validation loss should show a downward trend during training and converge or stabilize, rather than consistently increasing.

marianovitasari20:
I’m currently working on this.

marianovitasari20:
Changing the loss function can affect training and validation loss behavior. For example, in fully deterministic mode for a fair comparison, I compared: 
(1) L1 (MAE) loss: https://wandb.ai/turing-seaice/train/runs/4zdg0p8i (the loss increases)
(2) MSE loss: https://wandb.ai/turing-seaice/train/runs/v0u37f1z (the loss decreases)
(3) Weighted MSE loss (with weights from the active grid cell mask): 

I need to run more experiments with additional epochs and different seeds before drawing conclusions. I haven’t finished yet. Isambard has just come back online, it was down yesterday.

marianovitasari20:
1) L1 (MAE) loss

<img width="966" height="322" alt="Image" src="https://github.com/user-attachments/assets/829dce45-5af3-45ec-8885-3e09c966cb66" />

2) MSE loss

<img width="964" height="318" alt="Image" src="https://github.com/user-attachments/assets/a1f862ed-13a7-49c8-8622-23e3dcffa4d5" />

marianovitasari20:
<img width="1055" height="312" alt="Image" src="https://github.com/user-attachments/assets/d2c567da-fb32-45ff-add4-362687081540" />

With fully deterministic mode (MAE loss), the validation loss decreases after reprojection instead of increasing. This does not happen for all seeds, so I need to run more tests with different seeds to confirm.

marianovitasari20:
MAE vs MSE

<img width="1163" height="313" alt="Image" src="https://github.com/user-attachments/assets/fb58ebbc-3fb7-4d96-b266-ef2c0c59d976" />

<img width="1159" height="314" alt="Image" src="https://github.com/user-attachments/assets/8bf9a773-f192-42e8-a76a-6bcfc6b23003" />

marianovitasari20:
seed: 555
<img width="1162" height="317" alt="Image" src="https://github.com/user-attachments/assets/db0f3bf9-4fd4-422b-8b0e-88353adaf796" />
[...]

PR #261, opened 2026-04-29
256 fix seed

marianovitasari20:


marianovitasari20:
I’ve merged the latest changes from main and accepted the main version due to conflicts. Since the function I modified in this branch was changed or deleted in main, I now need to reapply my own changes, so I’m working on it again.

icenet_mp/model_service.py
```diff
@@ -19,12 +21,41 @@
 log = logging.getLogger(__name__)
 
 
+class _DeterministicInterpolate:
+    """Monkey-patch F.interpolate to strip antialias=True for deterministic CUDA backward.
```
jemrobinson, line 25: I'm not sure this is something we want to do. We previously needed the antialias argument to avoid artifacts when resizing.

icenet_mp/model_service.py
```diff
@@ -19,12 +21,41 @@
...
+
 class ModelService:
     def __init__(self, config: DictConfig) -> None:
         """Initialize the model service."""
         self.config_ = config
         if seed := config.get("seed", None):
-            seed_everything(int(seed), workers=True)
+            seed = int(seed)
+            os.environ["PYTHONHASHSEED"] = str(seed)
+            os.environ["CUBLAS_WORKSPACE_CONFIG"] = ":4096:8"
```
jemrobinson, line 59: Reference https://docs.nvidia.com/cuda/cublas/ and explain what this means

marianovitasari20:
It’s becoming nondeterministic again after I merge `main` into this branch. I’ve been investigating it since yesterday. I’ve only run it locally and haven’t tried it on Isambard yet.

jemrobinson (changes requested):
I'm not totally convinced that this is a thing we want to be doing.

If we believe that the real performance of a model training has quite a wide distribution of performance (i.e. a large spread of values across any particular metric) then what do we achieve if we manage to make this deterministic? If we want to compare "model configuration A" to "model configuration B" we would still need to run each of these 10-20 times with different seeds in order to convince ourselves that we understand the model uncertainty.

Would we be better working on running model ensembles instead?

IFenton:
> If we want to compare "model configuration A" to "model configuration B" we would still need to run each of these 10-20 times with different seeds in order to convince ourselves that we understand the model uncertainty.

My understanding was that it would allow us to compare two model configurations better, e.g. if model config A always has a lower error than model config B when they are run with the same seed then we can be more confident that model config A is a sensible thing to do. You'd still need to run each pairing a number of times with different seeds to get more of an idea of spread (/do model ensembles), but it seems like it allows a more like-for-like comparison of two configurations. Is that logic wrong @jemrobinson?

> Would we be better working on running model ensembles instead?
This feels like something we should be doing going forward, but as well rather instead.

marianovitasari20:
> I'm not totally convinced that this is a thing we want to be doing.
> 
> If we believe that the real performance of a model training has quite a wide distribution of performance (i.e. a large spread of values across any particular metric) then what do we achieve if we manage to make this deterministic? If we want to compare "model configuration A" to "model configuration B" we would still need to run each of these 10-20 times with different seeds in order to convince ourselves that we understand the model uncertainty.
> 
> Would we be better working on running model ensembles instead?

I think these are two separate things, @jemrobinson. First, we want to fairly investigate why a particular method behaves the way it does, and for that we need a fair, controlled comparison to draw meaningful conclusions. The deterministic mode with a fixed seed is just a tool for that, it doesn't remove uncertainty from the model itself.
Regarding ensembles, we can absolutely still do that later. Adding a seed here doesn't prevent us from running ensemble experiments, the seed is just for controlled comparisons during development. Under normal settings, we wouldn't use a seed at all.

Yes, I’m aware of the effect of removing antialiasing. I thought we could still use this to compare runs in a fully deterministic mode, and we wouldn't need to use a seed under normal settings. I take your point that ensembles could also be used for uncertainty quantification, and that they can be applied under normal settings without seeds later. However, I’m still inclined to keep the deterministic approach, as it makes it easier to compare different configurations, such as methods or loss functions. The purpose is to ensure a fair comparison. We will revert to running without a seed once we have decided which methods (e.g., loss functions or approaches) is better.

jemrobinson:
Hi @marianovitasari20. I guess my underlying point is this: what does comparing two different configurations with the same seed actually tell us? Each one is a single draw from an underlying distribution with large variance, so we can't simply compare e.g. MAE for the two configurations and say that one is performing better than the other.

What's the thing you want to compare that is made easier by this?

jemrobinson:
> My understanding was that it would allow us to compare two model configurations better, e.g. if model config A always has a lower error than model config B when they are run with the same seed then we can be more confident that model config A is a sensible thing to do. You'd still need to run each pairing a number of times with different seeds to get more of an idea of spread (/do model ensembles), but it seems like it allows a more like-for-like comparison of two configurations. Is that logic wrong @jemrobinson?

Yes, if model A always has lower error than model B across N runs, then we can conclude that it's probably performing better. However, I think this applies whether or not you use the same seed.

marianovitasari20:
> > My understanding was that it would allow us to compare two model configurations better, e.g. if model config A always has a lower error than model config B when they are run with the same seed then we can be more confident that model config A is a sensible thing to do. You'd still need to run each pairing a number of times with different seeds to get more of an idea of spread (/do model ensembles), but it seems like it allows a more like-for-like comparison of two configurations. Is that logic wrong @jemrobinson?
> 
> Yes, if model A always has lower error than model B across N runs, then we can conclude that it's probably performing better. However, I think this applies whether or not you use the same seed.

I understand your point, that’s why we want to run A vs. B with the same seed, and then repeat with different seeds. This still gives multiple runs, but in a fairer way.

marianovitasari20:
Just as an example, I ran these experiments using the same seed (while still in fully deterministic mode, before merging main into this branch) to compare the effect of using L1 loss vs. weighted MSE loss. I still need to run multiple experiments with different seeds before drawing conclusions later, but this is a fairer way.
https://github.com/alan-turing-institute/icenet-mp/issues/258

jemrobinson:
I don't quite see why it's fairer to compare different models with the same seed rather than different seeds. They're going to be drawing a different number of random numbers and using them for different purposes, so I think you're still comparing a single sample from a distribution of trained models against a single sample from a different distribution of trained models.

marianovitasari20:
Discussed further during the coworking meeting with @IFenton and @jemrobinson today.

marianovitasari20:
Removing _DeterministicInterpolate, the team agreed not to go with this approach. Keeping a note here for reference: if anyone wants fully deterministic behaviour in future, this function shows one way to do it. It does come with a performance cost, though personally I think it's acceptable for comparison purposes.

jemrobinson:
> Removing _DeterministicInterpolate, the team agreed not to go with this approach. Keeping a note here for reference: if anyone wants fully deterministic behaviour in future, this function shows one way to do it. It does come with a performance cost, though personally I think it's acceptable for comparison purposes.

To be clear - performance cost isn't a problem, but removing the "antialiasing" argument means that the behaviour of the model is changed.

marianovitasari20:
> > Removing _DeterministicInterpolate, the team agreed not to go with this approach. Keeping a note here for reference: if anyone wants fully deterministic behaviour in future, this function shows one way to do it. It does come with a performance cost, though personally I think it's acceptable for comparison purposes.
> 
> To be clear - performance cost isn't a problem, but removing the "antialiasing" argument means that the behaviour of the model is changed.

Sorry for the confusion, by "performance cost" I meant model quality, not compute. Stripping antialias changes the interpolation behaviour which could affect results. So yes, we're agreeing on the same thing.

marianovitasari20:
I've removed it now.

antialias=True → better signal processing, less aliasing, but nondeterministic
antialias=False → slightly worse interpolation quality, but deterministic

In most ML cases, the difference in quality is negligible compared to reproducibility and stable debugging. 

After more testing, I’m actually still not fully convinced we should remove this, but we can merge it for now rather than delay things further due to disagreement.

jemrobinson:
Can you add some plots/metrics here for discussion?

marianovitasari20:
> Can you add some plots/metrics here for discussion?

Sorry, I just realised it can’t write to Isambard due to permission issues, so I can’t access the checkpoints. I’ll come back to this once that’s been sorted out.

marianovitasari20:
https://wandb.ai/turing-seaice/evaluate/runs/wltxre91 (without _DeterministicInterpolate, antialias=True)
https://wandb.ai/turing-seaice/evaluate/runs/70evz1gf (with _DeterministicInterpolate, antialias=False)
comparison: https://wandb.ai/turing-seaice/evaluate/panel/f6pv3rhsg

jemrobinson:
@marianovitasari20 : can you attach a comparison plot to this issue instead of just the W&B links? Also, if the purpose of these plots are to demonstrate the effect of deterministic running, could we see:

- 5 runs using the same seed fully-deterministic
- 5 runs using the same seed but without _DeterministicInterpolate
- 5 runs using different seeds

I think that would help to show that deterministic running is removing or reducing the spread in the models.

marianovitasari20:
**Training plots**
1) 5 runs using the same seed fully-deterministic (seed: 123) --> the results won’t vary, so this is useful for comparison.
image
2) 5 runs using the same seed but without _DeterministicInterpolate (seed: 123)
image
3) 5 runs using different seeds (seeds: 17, 482, 903, 256, 771)
image

marianovitasari20:
**Point 1 (fully-deterministic) vs Point 3 (non-deterministic)**

seed: 17
image

seed: 482
image

seed: 903
image

seed: 256
image

seed: 771
image

marianovitasari20:
additional, 

seed: 771
image

seed: 256
image

marianovitasari20:
Any thoughts on this @jemrobinson @IFenton?

jemrobinson:
Can you show the spread of different seeds in the deterministic case on one graph and the spread in the non-deterministic case on another? No need for additional runs, but just a different way to see the data.

marianovitasari20:
> Can you show the spread of different seeds in the deterministic case on one graph and the spread in the non-deterministic case on another? No need for additional runs, but just a different way to see the data.

thanks @jemrobinson. I reran two of the seeds (without deterministic function) just to show that it keeps changing. Do you mean like this?

with DET:
image

without DET:
image

jemrobinson:
Yes, that's exactly the right thing. I think this confirms that there's quite a lot of spread between different runs that we can only capture by running multiple different seeds.

Given this, is there still something that a fully-deterministic run would be useful for? If there is, perhaps this is something we want to enable/disable in the base config?

icenet_mp/model_service.py
```diff
@@ -184,6 +190,20 @@ def build_trainer(
...
+
+        # Check warn_only survived Lightning's deterministic setup
+        log.debug(
+            "deterministic_algorithms_enabled: %s",
+            torch.are_deterministic_algorithms_enabled(),
+        )
+        log.debug(
+            "warn_only_enabled: %s",
+            torch.is_deterministic_algorithms_warn_only_enabled(),
+        )
```
jemrobinson: I'm a little confused by this. What's the thing that might be overriden if deterministic=True and what are we doing here to reset it?

jemrobinson (dismissed):

icenet_mp/model_service.py
marianovitasari20: I wanted both `deterministic=True` and `warn_only=True`, but Lightning's Trainer with `deterministic=True` internally calls `torch.use_deterministic_algorithms(True)` without `warn_only`, overriding our earlier setting. Re-applying it here was my workaround, without it I was getting errors. Not 100% sure this is the cleanest way to do it though.

marianovitasari20:
> Yes, that's exactly the right thing. I think this confirms that there's quite a lot of spread between different runs that we can only capture by running multiple different seeds.
> 
> Given this, is there still something that a fully-deterministic run would be useful for? If there is, perhaps this is something we want to enable/disable in the base config?

Yes, I think deterministic runs are still useful. The key case is when you want to isolate the effect of a specific change (e.g. a new loss function) without seed variance confounding the comparison, a "better" result might just be a lucky seed rather than a genuine improvement. So I'd suggest to keep the deterministic code in place, but making it off by default, only enabled when you want controlled comparisons or need reproducibility for debugging. I can also add a warning in the config noting that enabling deterministic mode will disable antialiasing, so users are aware of the trade-off.

icenet_mp/model_service.py
jemrobinson: Can this be replaced by something simpler like:

```
if torch.are_deterministic_algorithms_enabled():
  torch.use_deterministic_algorithms(True, warn_only=True)
```

jemrobinson:
> > Yes, that's exactly the right thing. I think this confirms that there's quite a lot of spread between different runs that we can only capture by running multiple different seeds.
> > Given this, is there still something that a fully-deterministic run would be useful for? If there is, perhaps this is something we want to enable/disable in the base config?
> 
> Yes, I think deterministic runs are still useful. The key case is when you want to isolate the effect of a specific change (e.g. a new loss function) without seed variance confounding the comparison, a "better" result might just be a lucky seed rather than a genuine improvement. So I'd suggest to keep the deterministic code in place, but making it off by default, only enabled when you want controlled comparisons or need reproducibility for debugging. I can also add a warning in the config noting that enabling deterministic mode will disable antialiasing, so users are aware of the trade-off.

Fine with me. Do we want to separate "run with a given seed" from "run deterministically" in the config file?

marianovitasari20:
> > > Yes, that's exactly the right thing. I think this confirms that there's quite a lot of spread between different runs that we can only capture by running multiple different seeds.
> > > Given this, is there still something that a fully-deterministic run would be useful for? If there is, perhaps this is something we want to enable/disable in the base config?
> > 
> > 
> > Yes, I think deterministic runs are still useful. The key case is when you want to isolate the effect of a specific change (e.g. a new loss function) without seed variance confounding the comparison, a "better" result might just be a lucky seed rather than a genuine improvement. So I'd suggest to keep the deterministic code in place, but making it off by default, only enabled when you want controlled comparisons or need reproducibility for debugging. I can also add a warning in the config noting that enabling deterministic mode will disable antialiasing, so users are aware of the trade-off.
> 
> Fine with me. Do we want to separate "run with a given seed" from "run deterministically" in the config file?

Do you mean to put the seed and deterministic settings separately in the config, like in the beginning?
like this:
seed: 123
fully_deterministic: True # this will turn off anti-aliasing

IFenton:
> Yes, I think deterministic runs are still useful. The key case is when you want to isolate the effect of a specific change (e.g. a new loss function) without seed variance confounding the comparison, a "better" result might just be a lucky seed rather than a genuine improvement. So I'd suggest to keep the deterministic code in place, but making it off by default, only enabled when you want controlled comparisons or need reproducibility for debugging. I can also add a warning in the config noting that enabling deterministic mode will disable antialiasing, so users are aware of the trade-off.

I agree this sounds like a good solution

jemrobinson:
> Do you mean to put the seed and deterministic settings separately in the config, like in the beginning? like this: seed: 123 fully_deterministic: True # this will turn off anti-aliasing

Yes, but probably grouped together under a heading, so something like:

```
random:
  fully_deterministic: True # this will turn off anti-aliasing
  seed: 123
```

and, in fact, probably best to put these in a new directory under `config/` with a `default.yaml` that doesn't have a seed or deterministic running, and a `deterministic.yaml` that does (see e.g. how `predict` has multiple different configs nested under it).

marianovitasari20:
Now, `seed` and `deterministic` are `null` and `false` by default. If you want to enable them, add the following to your `local.yaml`:

```yaml
random:
  fully_deterministic: true  
  seed: 123
```

icenet_mp/config/base.yaml
```diff
@@ -6,9 +6,8 @@ defaults:
...
   - model: naive_unet_naive
   - predict: sic-icenet-2d
+  - random: default
   - train: default
   - _self_
 
-base_path: ../base
-
-seed: 555
+base_path: /lus/lfs1aip2/projects/public/u6iz/shared/base
```
IFenton: Did you need to change this base path, given you change it for base_isambardai.yaml?

IFenton (approved):
Generally LGTM (and runs as expected). Just a minor comment about the base_path

icenet_mp/config/base.yaml
marianovitasari20: Thanks @IFenton. Sorry, yes, you’re right. I’ve changed it back to the original one.

marianovitasari20:
Hi @jemrobinson,
Sorry, I dismissed the "changes requested" one (not sure how that works), because I got blocked from merging this, and I had already made changes based on the reviews. Based on our meeting today, we can merge it if everything is fine, and it has already been approved by @IFenton. Let me know if this is not the case. Thank you both for the reviews! :)

jemrobinson:
@marianovitasari20 one approval is definitely enough! Thanks for doing this :)

Issue #20, opened 2025-07-17
Generate minimal automated API documentation for the pipeline


aranas:
## Problem / Goal
We want a minimal, auto‑generated API site so every public module, class, and function is visible in one place and team members can easily discover current capabilities of the pipeline

## What good looks like

- [x] Docstrings defining inputs/outputs exist for all public functions/classes
- [x] Doc generator is configured
- [x] Build script — make docs (or tox -e docs) builds HTML locally.
- [ ] CI job — workflow docs-build.yml runs the build and fails on errors.
- [ ] Published output — generated site in docs/_site or GitHub Pages; link/badge added to README.md.
- [ ] Index page lists every module, class, method (signatures visible).
- [ ] Quick “How to update docs” note in CONTRIBUTING.md.

jemrobinson:
I can do this. Is this my task for the week?

jemrobinson:
Another vote for adding docs that cover the config files from #270.

PR #270, opened 2026-05-14
Project Argo float data onto EASE2 grid

### LINKED ISSUE #264: Project Argo to EASE grid
### Problem / Goal

We should project Argo float data on the EASE2 grid, not lat/lon.

### What does good look like?

- [x] Argo float data available on the EASE2 grid

### Related context

_No response_

jemrobinson:
Replaces the Argo float rectangular lat/lon interpolation grid with an EASE2 polar grid. The existing `grid_factory` logic is used to generate this from CRS+resolution+shape.

- `ArgoSource` now needs `crs`, `resolution`, `shape` instead of `grid_resolution_degrees`
- Updated `full_float*` and `samp_float*` configs to use the new `25p0km` datasets
- Tests updated for compatibility

compare_argo_2017-01-10

<details>
<summary>Same but for 2017-01-20</summary>
compare_argo_2017-01-20
</details>

<details>
<summary>Same but for 2017-01-30</summary>
compare_argo_2017-01-30
</details>

## Test runs
- Comparing data before reprojection to data after reprojection
- 1 month of training data
- 2 weeks of evaluation data
- 5 epochs
- `quick_test` model (`naive-unet-naive` with 128x128 latent space)
- 5 runs of `1p0-degree` (before reprojection) and 5 runs of ease2 (after reprojection)

<figure>
train_loss
validation_loss(1)
</figure>

Taking just the final value (after 5 epochs) we see:

| | Training loss | Validation loss |
| --- | --- | --- |
| 1p0 degree | 0.014 ± 0.001 | 0.234 ± 0.253 |
| EASE2 | 0.012 ± 0.002 | 0.069 ± 0.028 |

so a small but not statistically significant improvement by switching to the EASE2 grid.

**second test run using 10 runs each with deterministic seeds 1-10**

<figure>
train_loss
validation_loss
</figure>

Again taking just the final value after 5 epochs we see:

| | Training loss | Validation loss |
| --- | --- | --- |
| 1p0 degree | 0.012 ± 0.002 | 0.091 ± 0.049 |
| EASE2 | 0.012 ± 0.002 | 0.093 ± 0.039 |

and again no significant difference between the two.

Closes #264


icenet_mp/config/data/datasets/full_floatnorth_argo_25p0km_1999_2025_24h_v2.yaml
```diff
@@ -555,12 +555,20 @@ full-floatnorth-argo-1p0-1999-2025-24h-v1:
 
   input:
     pipe:
+      # Get the ARGO data
       - argo:
           area: "90/-180/0/180"  # northern hemisphere
-          grid_resolution_degrees: 1.0
+          crs: "EPSG:6931"
```
louisavz, line 561: For my own curiosity, why did we not grid it to 25km before, but ~111km?

icenet_mp/data_processors/sources/argo.py
```diff
@@ -28,20 +30,30 @@ def __init__(  # noqa: PLR0913
         context: Context,
         *,
         area: str,
+        crs: str,
         param: list[str],
-        ignore_missing_dates: bool = False,
-        grid_resolution_degrees: float = 1,
+        resolution: str,
```
louisavz, line 35: What's the benefit of keep resolution as str "25p0km" over an int or float?

icenet_mp/data_processors/sources/argo.py
```diff
@@ -28,20 +30,30 @@ def __init__(  # noqa: PLR0913
...
+        shape: tuple[int, int],
         distance_scale_km: float = 2000,
```
louisavz, line 37: Should this be an int?

louisavz (commented):
LGTM overall from following through the code and logics. I did not attempt to run the code, but I'm happy to do that Tuesday evening if you want me to.

icenet_mp/config/data/datasets/full_floatnorth_argo_25p0km_1999_2025_24h_v2.yaml
jemrobinson: I think @IFenton chose this as something that didn't take too long to calculate and that would rarely have two floats in the same grid cell.

icenet_mp/data_processors/sources/argo.py
jemrobinson: In the Anemoi file that we create, resolution is stored as a string. This is the string format for Lambert Azimuthal Equal Area resolutions that is used in Anemoi.

icenet_mp/data_processors/sources/argo.py
jemrobinson: I think it could be, but we're only using it in `weights = np.exp(-0.5 * (distance_km / self.distance_scale_km) ** 2)` so it will probably be converted to some kind of numpy float here anyway.

marianovitasari20:
currently reviewing

marianovitasari20:
Can you show the impact of this feature in the WandB run, with and without it enabled?

marianovitasari20:
Could we add (or update) documentation covering how to configure the Argo data source for the different opt-in/opt-out scenarios? Specifically:
* With vs. without Argo: what does a minimal config look like when Argo is excluded entirely vs. included?
* With vs. without the EASE2 grid: is there a supported path to skip the EASE2 projection, or is it now always required?

Since `ArgoSource` now requires `crs`, `resolution`, and `shape` instead of the old `grid_resolution_degrees`, it would be helpful to have a reference showing the expected config structure for each setup.

The new dataset YAML files are a good implicit example, but explicit docs (even a short README section or inline comments in an example config) would make it much easier for others to understand and use correctly.

jemrobinson:
> Could we add (or update) documentation covering how to configure the Argo data source for the different opt-in/opt-out scenarios? Specifically:

I think that we definitely **should** have better documentation of how to do things like this, but this PR probably isn't the right place for it. Our current docs are very minimal and just live in the README. We really need to do #20.
 
> With vs. without Argo: what does a minimal config look like when Argo is excluded entirely vs. included?

Here's our default `sample` dataset config:

```
defaults:
  - datasets:
    - samp_floatsouth_argo_25p0km_2017_2019_24h_v2
    - samp_sicsouth_osisaf_25p0km_2017_2019_24h_v2
    - samp_weathersouth_era5_25p0km_2017_2019_24h_v3
  - split: sample_dataset
  - _self_
```

so running without Argo would simply be removing this line (or overriding the `/data/datasets` key)

> With vs. without the EASE2 grid: is there a supported path to skip the EASE2 projection, or is it now always required?

Yes, you can either use the 1.0 degree dataset (e.g. `samp_floatsouth_argo_1p0_2017_2019_24h_v1.yaml`) or the EASE2 grid dataset (e.g. `samp_floatsouth_argo_25p0km_2017_2019_24h_v2.yaml`).

> Since `ArgoSource` now requires `crs`, `resolution`, and `shape` instead of the old `grid_resolution_degrees`, it would be helpful to have a reference showing the expected config structure for each setup.
>
> The new dataset YAML files are a good implicit example, but explicit docs (even a short README section or inline comments in an example config) would make it much easier for others to understand and use correctly.

Is this question about creating a new dataset config from scratch?  Are you thinking that we might need more datasets than the following:

- `samp_floatsouth_argo_1p0_2017_2019_24h_v2.yaml` and `samp_floatsouth_argo_25p0km_2017_2019_24h_v2.yaml` for quick, local testing
- `full_floatsouth_argo_25p0km_1999_2025_24h_v12.yaml` and full_floatnorth_argo_25p0km_1999_2025_24h_v12.yaml` for full runs

If it's a question about "how do we write a dataset config" I'll add a note to #20.

jemrobinson:
> Can you show the impact of this feature in the WandB run, with and without it enabled?

Done and added to the top-level comment

marianovitasari20:
Thanks for the explanation @jemrobinson! For the full documentation, I agree that's better handled via https://github.com/alan-turing-institute/icenet-mp/issues/20. But could we still add, even just a brief note in the existing README or inline comments in the example configs, e.g. "remove this line to exclude Argo", or noting which dataset YAML to use for EASE2 or other grid, as a small addition to this PR for clarity?

marianovitasari20:
> > Can you show the impact of this feature in the WandB run, with and without it enabled?
> 
> Done and added to the top-level comment

Thanks. Do you have a comparison using the same seeds across multiple runs?

Thanks for the test runs! Could you confirm whether these were run with a fixed seed or not? If not, would it be possible to rerun with the same seed across both conditions to reduce variance (via multiple runs/seeds)?

jemrobinson:
> Thanks for the explanation @jemrobinson! For the full documentation, I agree that's better handled via #20. But could we still add, even just a brief note in the existing README or inline comments in the example configs, e.g. "remove this line to exclude Argo", or noting which dataset YAML to use for EASE2 or other grid, as a small addition to this PR for clarity?

Done in 8b56ef06

marianovitasari20:
> > Thanks for the explanation @jemrobinson! For the full documentation, I agree that's better handled via #20. But could we still add, even just a brief note in the existing README or inline comments in the example configs, e.g. "remove this line to exclude Argo", or noting which dataset YAML to use for EASE2 or other grid, as a small addition to this PR for clarity?
> 
> Done in [8b56ef0](https://github.com/alan-turing-institute/icenet-mp/commit/8b56ef06f6512772e6b5ab9bee5571746d1e43f7)

LGTM!

marianovitasari20 (commented):
LGTM. Once the deterministic results finish, I’ll approve this PR.

README.md
```diff
@@ -106,9 +104,39 @@ uv run imp datasets create --config-name <a config file that requires this datas
...
+To understand how dataset properties are encoded in its name, read `config/data/datasets/naming_convention.txt`.
+To define a custom set of datasets, for example for comparison testing, you can do the following.
+
+- create a custom dataset list in `config/data/my_datasets.local.yaml` that might look like this:
+
+```
+defaults:
+  - datasets:
+    - samp_sicsouth_osisaf_25p0km_2017_2019_24h_v2
+    - samp_weathersouth_era5_0p5_2017_2019_24h_v3
```
marianovitasari20: Sorry, I just realized, is it actually v3 or v2?

marianovitasari20: I mean: samp_weathersouth_era5_0p5_2017_2019_24h_v3 or samp_weathersouth_era5_0p5_2017_2019_24h_v2?

marianovitasari20:
Sorry, I’m a bit confused by the new figures. Could you please provide a comparison for each matching seed?

README.md
jemrobinson: Yes, this is fixed in abac70c2.

jemrobinson:
> Sorry, I’m a bit confused by the new figures. Could you please provide a comparison for each matching seed?

Like this?

**seed = 1**
train_loss_1

**seed = 2**
train_loss_2

I'm not sure what this tells us though.

marianovitasari20:
> I'm not sure what this tells us though.

yes, exactly!
By keeping the seed fixed, we isolate the effect of the change itself, any difference we see is due to the modification, not randomness. Once we've compared each seed that way, we can then look across seeds to see how consistent that effect is.

From these two figures, the EASE2 grid looks like it provides a small improvement.
Could you please include the rest of the other seeds as well?

jemrobinson:
> yes, exactly! By keeping the seed fixed, we isolate the effect of the change itself, any difference we see is due to the modification, not randomness. Once we've compared each seed that way, we can then look across seeds to see how consistent that effect is.

That might be true in this case, since we're using a 0-parameter deterministic encoder, but I don't think it's true in general. For a non-trivial encoder, backprop through a [90, 360] -> [128, 128] encoder is going to use a different number of random numbers than backprop through a [432, 432] -> [128, 128] encoder so as soon as you've run a single timestep, the remaining backpropagation through the rest of the model will be starting from a different place in the pseudo-random number stream which is equivalent to using different random numbers.
 
> From these two figures, the EASE2 grid looks like it provides a small improvement. Could you please include the rest of the other seeds as well?

Here they are training (left) and validation (right) loss for each seed from 3-10. In some cases EASE2 is better than native grid and in some cases it's worse.

I think that the mean/stddev across the ensemble which I show in the top-comment is more informative.

<figure>
train_3
valid_3
train_4
valid_4
train_5
valid_5
valid_6
train_6
train_7
valid_7
train_8
valid_8
train_9
valid_9
train_10
valid_10
</figure>

marianovitasari20 (approved):

PR #283, opened 2026-05-28
Make loss function configurable via Hydra config

### LINKED ISSUE #53: Make loss configurable
### Problem / Goal

Loss function is currently hard-coded

### What does good look like?

- [ ] Multiple loss functions available
- [ ] Config file allows choice of loss function to be specified

### Related context

_No response_

### LINKED ISSUE #173: What is our default loss function?
### What I need answered

What loss function does our pipeline default to and how does that compare to original icenet paper?

### Why it matters

Important information to know when interpreting model behavior

### What cannot happen without this answer?

- analysis of viT performance that uses default loss

### Who can help?

_No response_

marianovitasari20:
This PR adds support for configuring the loss function (MSE, MAE, RMSE, Huber, Smooth L1) through the Hydra config and updates BaseModel to instantiate the loss from config. 

The default loss is set to Huber in base.yaml as it generally performs best for this use case.

Closes #53. Closes #173.

WandB Huber: https://wandb.ai/turing-seaice/train/runs/8y9hnnwh


icenet_mp/models/base_model.py
```diff
@@ -77,16 +76,23 @@ def __init__(  # noqa: PLR0913
...
+
+        if loss:
+            self._loss_fn: torch.nn.Module = hydra.utils.instantiate(loss)
+        else:
+            log.warning(
+                "No loss config provided to %s; falling back to torch.nn.MSELoss(). "
+                "Check that your Hydra config mounts a loss at `model.loss`.",
+                name,
+            )
+            self._loss_fn = torch.nn.MSELoss()
```
jemrobinson: I don't think the if/else is needed here. We're assuming that a valid config file is being passed and allowing hydra to raise initialisation exceptions if it's not (see e.g. scheduler in the __init__ for this class).

```suggestion
        self._loss_fn: torch.nn.Module = hydra.utils.instantiate(loss)
```

icenet_mp/models/base_model.py
```diff
@@ -42,6 +40,7 @@ def __init__(  # noqa: PLR0913
         optimizer: DictConfig,
         output_space: DictConfig,
         scheduler: DictConfig,
+        loss: DictConfig | None = None,
```
jemrobinson: Let's not allow an empty argument here.

```suggestion
        loss: DictConfig,
```

jemrobinson (changes requested):
Looks good. We also need a `loss=...` argument to the `hydra.utils.instantiate` call in `ModelService.from_config` to allow the config argument to be passed through to the `BaseModel`.

jemrobinson:
Can you make the PR title a bit more descriptive and add some detail to the summary comment (including that this closes issue 53)?

icenet_mp/models/base_model.py
marianovitasari20: Got it, so the default should live in the config (e.g. base.yaml) rather than being handled in code here. I'll remove the if/else. I assume if loss is missing from the config, Hydra will raise a missing value error?

marianovitasari20:
> Looks good. We also need a `loss=...` argument to the `hydra.utils.instantiate` call in `ModelService.from_config` to allow the config argument to be passed through to the `BaseModel`.

thanks @jemrobinson. I think this may already be handled. Hydra's instantiate automatically passes all keys from the config to the constructor, so loss is already being passed through from cfg.model without needing to add it explicitly. I've tested it and it works. Happy to add it explicitly if you'd prefer?

jemrobinson:
> > Looks good. We also need a `loss=...` argument to the `hydra.utils.instantiate` call in `ModelService.from_config` to allow the config argument to be passed through to the `BaseModel`.
> 
> thanks @jemrobinson. I think this may already be handled. Hydra's instantiate automatically passes all keys from the config to the constructor, so loss is already being passed through from cfg.model without needing to add it explicitly. I've tested it and it works. Happy to add it explicitly if you'd prefer?

Interesting. I would think that this section (https://github.com/alan-turing-institute/icenet-mp/blob/main/icenet_mp/model_service.py#L80-L93) means we're only passing the `model` part of the config plus the other arguments that are passed explicitly.

Can you add some tests that changing the loss function in the config changes it in the model?

icenet_mp/config/losses/huber.yaml
```diff
@@ -0,0 +1,32 @@
...
+#     - Larger delta  → transition later,  more MSE-like (smoother near zero)
+#     - PyTorch default: 1.0
+#
+# When to use:
+#   - Regression where outliers would otherwise dominate an MSE loss
+#   - When you want the loss value to stay in the same units as the target
+# =============================================================================
+
+_target_: torch.nn.HuberLoss
+delta: 0.5  #0.0 to 1.0
```
jemrobinson: What does this comment mean?

marianovitasari20:
> > > Looks good. We also need a `loss=...` argument to the `hydra.utils.instantiate` call in `ModelService.from_config` to allow the config argument to be passed through to the `BaseModel`.
> > 
> > 
> > thanks @jemrobinson. I think this may already be handled. Hydra's instantiate automatically passes all keys from the config to the constructor, so loss is already being passed through from cfg.model without needing to add it explicitly. I've tested it and it works. Happy to add it explicitly if you'd prefer?
> 
> Interesting. I would think that this section (https://github.com/alan-turing-institute/icenet-mp/blob/main/icenet_mp/model_service.py#L80-L93) means we're only passing the `model` part of the config plus the other arguments that are passed explicitly.
> 
> Can you add some tests that changing the loss function in the config changes it in the model?

Do you mean something like this? https://wandb.ai/turing-seaice/train/runs/8y9hnnwh/files/model_config.yaml

icenet_mp/config/losses/huber.yaml
marianovitasari20: I've added more comments to the documentation, done in https://github.com/alan-turing-institute/icenet-mp/pull/283/changes/37b16f16c5afa7ac42d4e3a6dcde172ab51f83dc

icenet_mp/models/base_model.py
marianovitasari20: I just realised, after removing the if/else, the CI failed because Persistence doesn't use a loss function but now inherits the required loss argument from BaseModel. I could fix it by popping loss from kwargs in Persistence, but if we add more models without a loss in future we'd have to do the same thing each time. I think keeping the if/else in BaseModel is more robust. What do you think, or do you have a preferred approach?

marianovitasari20:
https://github.com/alan-turing-institute/icenet-mp/pull/283#discussion_r3333159382

I just realised, after removing the if/else, the CI failed because Persistence doesn't use a loss function but now inherits the required loss argument from BaseModel. I could fix it by popping loss from kwargs in Persistence, but if we add more models without a loss in future we'd have to do the same thing each time. I think keeping the if/else in BaseModel is more robust. What do you think, or do you have a preferred approach @jemrobinson?

jemrobinson:
> > Can you add some tests that changing the loss function in the config changes it in the model?
> 
> Do you mean something like this? https://wandb.ai/turing-seaice/train/runs/8y9hnnwh/files/model_config.yaml

I mean something in the `tests/` folder that initialises a model from a config and checks:

1. that the model can't be initialised without a loss section
2. that initialising with loss function X creates a model that uses loss function X (for each X)
3. that initialising with a loss function that doesn't exist fails

jemrobinson:
> I just realised, after removing the if/else, the CI failed because Persistence doesn't use a loss function but now inherits the required loss argument from BaseModel. I could fix it by popping loss from kwargs in Persistence, but if we add more models without a loss in future we'd have to do the same thing each time. I think keeping the if/else in BaseModel is more robust. What do you think, or do you have a preferred approach @jemrobinson?

I'm confused about why this is a problem. The `persistence` config inherits from `base` which has a default `losses` argument, right? So that should get passed to the `BaseModel` and used to build `self._loss_fn`. Is persistence being called in a way that doesn't pass a `losses` config?

marianovitasari20:
> > I just realised, after removing the if/else, the CI failed because Persistence doesn't use a loss function but now inherits the required loss argument from BaseModel. I could fix it by popping loss from kwargs in Persistence, but if we add more models without a loss in future we'd have to do the same thing each time. I think keeping the if/else in BaseModel is more robust. What do you think, or do you have a preferred approach @jemrobinson?
> 
> I'm confused about why this is a problem. The `persistence` config inherits from `base` which has a default `losses` argument, right? So that should get passed to the `BaseModel` and used to build `self._loss_fn`. Is persistence being called in a way that doesn't pass a `losses` config?

I think the Python test failure is because the tests instantiate `persistence` directly without loss (since persistence doesn't need loss). This worked before because loss had a default of None and the if/else handled it. Now that both are removed, the tests fail. Should I revert both changes, putting back `loss: DictConfig | None = None` and the if/else fallback? It works well with both and the tests pass.

jemrobinson:
> I think the Python test failure is because the tests instantiate `persistence` directly without loss (since persistence doesn't need loss). This worked before because loss had a default of None and the if/else handled it. Now that both are removed, the tests fail. Should I revert both changes, putting back `loss: DictConfig | None = None` and the if/else fallback? It works well with both and the tests pass.

Looking at this test failure

```
self = <test_base_model.TestBaseModel object at 0x7fdb1f94d450>
cfg_input_space = {'channels': 4, 'name': 'test-input', 'shape': [512, 512]}
cfg_output_space = {'channels': 1, 'name': 'target', 'shape': [432, 432]}

    def test_loss(
        self, cfg_input_space: DictConfig, cfg_output_space: DictConfig
    ) -> None:
        model = FakeDataModel(
            name="fake data",
            input_spaces=[cfg_input_space],
            n_forecast_steps=1,
            n_history_steps=1,
            output_space=cfg_output_space,
            optimizer=DictConfig({}),
            scheduler=DictConfig({}),
        )
        # Test loss
        prediction = torch.zeros(1, 1, 1, 1)
        target = torch.ones(1, 1, 1, 1)
>       assert model.loss(prediction, target) == torch.tensor(1.0)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

it's because our `FakeDataModel` isn't created with a `losses` argument. We should add one to the test.

marianovitasari20:
> > I think the Python test failure is because the tests instantiate `persistence` directly without loss (since persistence doesn't need loss). This worked before because loss had a default of None and the if/else handled it. Now that both are removed, the tests fail. Should I revert both changes, putting back `loss: DictConfig | None = None` and the if/else fallback? It works well with both and the tests pass.
> 
> Looking at this test failure
> 
> ```
> self = <test_base_model.TestBaseModel object at 0x7fdb1f94d450>
> cfg_input_space = {'channels': 4, 'name': 'test-input', 'shape': [512, 512]}
> cfg_output_space = {'channels': 1, 'name': 'target', 'shape': [432, 432]}
> 
>     def test_loss(
>         self, cfg_input_space: DictConfig, cfg_output_space: DictConfig
>     ) -> None:
>         model = FakeDataModel(
>             name="fake data",
>             input_spaces=[cfg_input_space],
>             n_forecast_steps=1,
>             n_history_steps=1,
>             output_space=cfg_output_space,
>             optimizer=DictConfig({}),
>             scheduler=DictConfig({}),
>         )
>         # Test loss
>         prediction = torch.zeros(1, 1, 1, 1)
>         target = torch.ones(1, 1, 1, 1)
> >       assert model.loss(prediction, target) == torch.tensor(1.0)
>                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
> ```
> 
> it's because our `FakeDataModel` isn't created with a `losses` argument. We should add one to the test.


I've gone with passing `loss=_DEFAULT_LOSS` to `Persistence` in the tests (https://github.com/alan-turing-institute/icenet-mp/pull/283/changes/0bc05429cab27aecabe88186344a28509d59afd8) to keep things consistent and get CI passing. Worth noting though that this carries a small risk. If we add more models in the future that genuinely don't need a loss (like `Persistence`), each one will need to either receive a loss it doesn't use, or we'll need to revisit making `loss` optional in `BaseModel`.

marianovitasari20:
> > > Can you add some tests that changing the loss function in the config changes it in the model?
> > 
> > 
> > Do you mean something like this? https://wandb.ai/turing-seaice/train/runs/8y9hnnwh/files/model_config.yaml
> 
> I mean something in the `tests/` folder that initialises a model from a config and checks:
> 
> 1. that the model can't be initialised without a loss section
> 2. that initialising with loss function X creates a model that uses loss function X (for each X)
> 3. that initialising with a loss function that doesn't exist fails

Currently doing this, will let you know when I’m finished.

marianovitasari20:
> > > > Can you add some tests that changing the loss function in the config changes it in the model?
> > > 
> > > 
> > > Do you mean something like this? https://wandb.ai/turing-seaice/train/runs/8y9hnnwh/files/model_config.yaml
> > 
> > 
> > I mean something in the `tests/` folder that initialises a model from a config and checks:
> > 
> > 1. that the model can't be initialised without a loss section
> > 2. that initialising with loss function X creates a model that uses loss function X (for each X)
> > 3. that initialising with a loss function that doesn't exist fails
> 
> Currently doing this, will let you know when I’m finished.

done in https://github.com/alan-turing-institute/icenet-mp/pull/283/changes/99209348babff5adfc0e04c06f8e2d7c7aba01b7, https://github.com/alan-turing-institute/icenet-mp/pull/283/changes/8ea51c5f6c91573bbfa5633f263784a63a715178, https://github.com/alan-turing-institute/icenet-mp/pull/283/changes/deb3789db95b8ab869016882477965727e4a27f8

marianovitasari20:
Hi @jemrobinson, are you happy with the changes, or do you have any additional feedback?

icenet_mp/config/loss/rmse.yaml
```diff
@@ -0,0 +1,2 @@
+# @package model.loss
```
jemrobinson: If we're going to document our loss functions (as you've done for the others) can you do so here too?

icenet_mp/models/base_model.py
```diff
@@ -1,7 +1,7 @@
 import itertools
+import logging
```
jemrobinson: Do we need this (and the subsequent `log = logging.getLogger(__name__)`)?

icenet_mp/config/base.yaml
```diff
@@ -8,6 +8,7 @@ defaults:
   - predict: sic-icenet-2d
   - random: default
   - train: default
+  - losses: huber
```
jemrobinson: I think this would be better as just `loss` with `icenet_mp/config/losses` renamed to `icenet_mp/config/loss` to match.

```suggestion
  - loss: huber
```

icenet_mp/config/loss/huber.yaml
```diff
@@ -0,0 +1,32 @@
+# @package model.loss
```
jemrobinson: This line means that the file contents will be included at config/model/loss regardless of how we specify the config structure.

This is interesting but I don't think it's what we want. It is better to specify loss as a top-level key (equivalently to `random`) since it's conceptually part of the way we *train* the model not part of the *architecture* of the model.

In order for this to work, we'll also need to add a line to `ModelService.from_config` that looks like `loss=config["loss"]`, as we do for e.g. `optimizer`.

tests/models/test_base_model.py
```diff
@@ -24,6 +24,9 @@ def forward(self, inputs: dict[str, TensorNTCHW]) -> TensorNTCHW:
         return torch.randn(b, self.t, self.c, self.h, self.w)
 
 
+_DEFAULT_LOSS = OmegaConf.create({"_target_": "torch.nn.HuberLoss", "delta": 0.5})
```
jemrobinson: This is fine, but it's also possible to put this into the definition of `FakeDataModel` so that it looks like

```
loss_cfg = OmegaConf.create({"_target_": "torch.nn.HuberLoss", "delta": 0.5})
super().__init__(*args, hemisphere="north", loss=loss_cfg, **kwargs)
```

and you then don't need all the `loss=_DEFAULT_LOSS` lines below.

jemrobinson (changes requested):
Overall this looks good - I just have a couple of suggestions about how to structure the config files.

Given the discussion on Slack - are we confident that we want to switch to Huber loss? Let's discuss in coworking.

icenet_mp/config/loss/rmse.yaml
marianovitasari20: done in https://github.com/alan-turing-institute/icenet-mp/pull/283/changes/a9b82c038570f712357b72fb9e8ad7b4b3fc2bf6

icenet_mp/models/base_model.py
marianovitasari20: You're right, we don't need them. Removed in https://github.com/alan-turing-institute/icenet-mp/pull/283/changes/db7c6bb2c5fcb5fd020512110d68e43cf4aea7de

icenet_mp/config/base.yaml
marianovitasari20: done in https://github.com/alan-turing-institute/icenet-mp/pull/283/changes/697eece6ccdefe405924ad0661e7c99b5150f77d

icenet_mp/config/loss/huber.yaml
marianovitasari20: done in https://github.com/alan-turing-institute/icenet-mp/pull/283/changes/70aafc1439f74b94458347dad8cc6eeb2099df97

tests/models/test_base_model.py
marianovitasari20: done in https://github.com/alan-turing-institute/icenet-mp/pull/283/changes/ee87d7c099315ea2b73ac0c560cf82d960d4eae2

marianovitasari20:
> Overall this looks good - I just have a couple of suggestions about how to structure the config files.
> 
> Given the discussion on Slack - are we confident that we want to switch to Huber loss? Let's discuss in coworking.

Thanks @jemrobinson for the reviews. From the experiments I've run so far, Huber loss has consistently shown better performance for our case. I haven't finished all the experiments yet, but I'd like to present what I have so far alongside the loss declines investigation results. Happy to discuss further in the meeting!

tests/models/test_base_model.py
```diff
@@ -112,7 +115,10 @@ def test_loss(
         # Test loss
         prediction = torch.zeros(1, 1, 1, 1)
         target = torch.ones(1, 1, 1, 1)
-        assert model.loss(prediction, target) == torch.tensor(1.0)
+        result = model.loss(prediction, target)
```
jemrobinson: Can we still check this against a specific Tensor? This should still be deterministic, right?

jemrobinson:
See also #304

marianovitasari20:
Please let me know if you're happy with these changes @jemrobinson. Thanks!

tests/models/test_loss_config.py
```diff
@@ -0,0 +1,98 @@
...
+                optimizer=DictConfig({}),
+                scheduler=DictConfig({}),
+                # loss intentionally omitted
+            )
+
+    # 2. Initialising with loss X creates a model that uses loss X
+    @pytest.mark.parametrize("loss_name", list(LOSS_CONFIGS.keys()))
+    @staticmethod
+    def test_loss_type(
+        loss_name: str,
```
jemrobinson, line 64:
```suggestion
    @pytest.mark.parametrize("loss_name", list(LOSS_CONFIGS.keys()))
    def test_loss_type(
        self,
        loss_name: str,
```
I don't think pytest deals well with staticmethods

tests/models/test_encode_process_decode.py
```diff
@@ -1,9 +1,11 @@
 import pytest
 import torch
-from omegaconf import DictConfig
+from omegaconf import DictConfig, OmegaConf
 
 from icenet_mp.models import EncodeProcessDecode
 
+_DEFAULT_LOSS = OmegaConf.create({"_target_": "torch.nn.HuberLoss", "delta": 0.5})
```
jemrobinson: This is repeated in a few places (test_encode_process_decode.py, test_persistence.py:43, test_base_model.py). If you instead write a fixture and put it in conftest.py we can just define it once.

icenet_mp/config/loss/smooth_l1.yaml
```diff
@@ -0,0 +1,30 @@
...
+#     - Smaller beta → transition sooner, more MAE-like (more robust to outliers)
+#     - Larger beta  → transition later,  more MSE-like (smoother near zero)
+#     - PyTorch default: 1.0
+#
+# When to use:
+#   - Regression where outliers would otherwise dominate an MSE loss
+#   - Popular in object detection (e.g. Fast R-CNN) for bounding-box regression
+# =============================================================================
+
+_target_: torch.nn.SmoothL1Loss
```
jemrobinson, line 29: This is very similar to HuberLoss. When would you use one over the other?

jemrobinson (changes requested):
Looks pretty good to me. A couple of minor comments before we merge.

jemrobinson:
@marianovitasari20 : can you also allow the functions we already have in the losses folder (WeightedBCEWithLogitsLoss, WeightedL1Loss and WeightedMSELoss) to be specified by the config file? They might need to be slightly adapted to handle the case where we don't pass sample weights.

icenet_mp/config/loss/smooth_l1.yaml
marianovitasari20: They're mathematically the same family, but PyTorch parameterizes them differently:

* `SmoothL1Loss(beta)`: the function value at the transition is `0.5 * beta`
* `HuberLoss(delta)`: the function value at the transition is `0.5 * delta²`

If `beta=1` (the default), they're numerically identical.

At the transition, the function value is `0.5 * delta²`. So `HuberLoss(delta) == delta * SmoothL1Loss(beta=delta)`. They produce identical gradients up to a constant scale factor of `delta`.

* **Smooth L1 Loss**: Smooth L1 Loss stays roughly the same size as the error itself, double the error, roughly double the loss (once you're outside the smooth region). The `beta` parameter controls the width of the smooth quadratic dip near zero: smaller `beta` means a narrower, more sharply curved dip. Historically used in Fast/Faster R-CNN bbox regression where `beta=1` is conventional. Good when you want the loss numerically comparable to L1.
* **Huber Loss**: The loss has two regimes: L2 (a curvy bowl) near zero, and L1 (straight lines) further out. In Huber Loss, delta controls where the handoff from L2 to L1 happens. Good when you have a sense of how big "normal" errors should be and want to clip the influence of anything beyond that. Standard in robust statistics and DQN-style RL.

tests/models/test_base_model.py
marianovitasari20: done in https://github.com/alan-turing-institute/icenet-mp/pull/283/changes/8bc129b2a515a8399344652e610047cfef1d6c23

tests/models/test_encode_process_decode.py
marianovitasari20: done in https://github.com/alan-turing-institute/icenet-mp/pull/283/changes/69794e0d443c0e9ed0287483960810cd4914d99d

marianovitasari20:
> @marianovitasari20 : can you also allow the functions we already have in the losses folder (WeightedBCEWithLogitsLoss, WeightedL1Loss and WeightedMSELoss) to be specified by the config file? They might need to be slightly adapted to handle the case where we don't pass sample weights.

done in https://github.com/alan-turing-institute/icenet-mp/pull/283/changes/a5b3220c1aa81d1f29e32ffdb7859c57029e2d48

icenet_mp/losses/weighted_l1_loss.py
```diff
@@ -20,19 +19,29 @@ def __init__(self, *args: Any, **kwargs: Any) -> None:
             **kwargs: Keyword arguments passed to torch.nn.L1Loss.
 
         """
+        kwargs["reduction"] = "none"
         super().__init__(*args, **kwargs)
```
jemrobinson, line 32:
```suggestion
        super().__init__(*args, reduction="none", **kwargs)
```

icenet_mp/losses/weighted_bce_loss.py
```diff
@@ -20,23 +20,29 @@ def __init__(self, *args: Any, **kwargs: Any) -> None:
             **kwargs: Keyword arguments passed to torch.nn.BCEWithLogitsLoss.
 
         """
+        kwargs["reduction"] = "none"
         super().__init__(*args, **kwargs)
```
jemrobinson, line 33:
```suggestion
        super().__init__(*args, reduction="none", **kwargs)
```

jemrobinson (changes requested):
LGTM. A couple of very small stylistics changes to consider.

marianovitasari20:
> LGTM. A couple of very small stylistics changes to consider.

Thanks @jemrobinson! I've applied your suggestions.

icenet_mp/losses/weighted_bce_loss.py
```diff
@@ -20,23 +20,28 @@ def __init__(self, *args: Any, **kwargs: Any) -> None:
             **kwargs: Keyword arguments passed to torch.nn.BCEWithLogitsLoss.
 
         """
-        super().__init__(*args, **kwargs)
+        super().__init__(*args, reduction="none", **kwargs)
```
marianovitasari20:
```suggestion
        kwargs["reduction"] = "none"
        super().__init__(*args, **kwargs)
```

icenet_mp/losses/weighted_l1_loss.py
```diff
@@ -20,19 +19,28 @@ def __init__(self, *args: Any, **kwargs: Any) -> None:
             **kwargs: Keyword arguments passed to torch.nn.L1Loss.
 
         """
-        super().__init__(*args, **kwargs)
+        super().__init__(*args, reduction="none", **kwargs)
```
marianovitasari20:
```suggestion
        kwargs["reduction"] = "none"
        super().__init__(*args, **kwargs)
```

marianovitasari20:
>> Changed the style back. Using super().__init__(*args, reduction="none", **kwargs) causes a "multiple values for keyword argument" error, so keeping kwargs["reduction"] = "none" before super().__init__ instead.

> Just to be clear, this means that you're passing `reduction` somewhere in the code (e.g. from the config file) but then ignoring it. Is this what you want?

We're overriding it intentionally. The weighted loss requires elementwise outputs (instead of scalar) before applying the weights, so "none" is the only valid value here.

jemrobinson:
> We're overriding it intentionally. The weighted loss requires elementwise outputs (instead of scalar) before applying the weights, so "none" is the only valid value here.

Do we instead want to raise an exception if the config is providing another value?

icenet_mp/losses/weighted_bce_loss.py
```diff
@@ -20,23 +20,29 @@ def __init__(self, *args: Any, **kwargs: Any) -> None:
             **kwargs: Keyword arguments passed to torch.nn.BCEWithLogitsLoss.
 
         """
```
marianovitasari20, line 25:
```suggestion
        """
        if "reduction" in kwargs and kwargs["reduction"] != "none":
        log.warning(
            "Ignoring reduction='%s'; this loss requires reduction='none' "
            "for weighted loss computation.", kwargs["reduction"]
         )
```

icenet_mp/losses/weighted_l1_loss.py
```diff
@@ -20,19 +19,29 @@ def __init__(self, *args: Any, **kwargs: Any) -> None:
             **kwargs: Keyword arguments passed to torch.nn.L1Loss.
 
         """
```
marianovitasari20, line 24:
```suggestion
        """
        if "reduction" in kwargs and kwargs["reduction"] != "none":
        log.warning(
            "Ignoring reduction='%s'; this loss requires reduction='none' "
            "for weighted loss computation.", kwargs["reduction"]
        )
```

icenet_mp/losses/weighted_mse_loss.py
marianovitasari20, line 25:
```suggestion
        """
        if "reduction" in kwargs and kwargs["reduction"] != "none":
        log.warning(
            "Ignoring reduction='%s'; this loss requires reduction='none' "
            "for weighted loss computation.", kwargs["reduction"]
        )
```

icenet_mp/losses/weighted_l1_loss.py
```diff
@@ -20,19 +19,34 @@ def __init__(self, *args: Any, **kwargs: Any) -> None:
...
+        if "reduction" in kwargs and kwargs["reduction"] != "none":
+        log.warning(
+            "Ignoring reduction='%s'; this loss requires reduction='none' "
+            "for weighted loss computation.", kwargs["reduction"]
+        )
```
marianovitasari20:
```suggestion
          log.warning(
              "Ignoring reduction='%s'; this loss requires reduction='none' "
              "for weighted loss computation.", kwargs["reduction"]
          )
```

icenet_mp/losses/weighted_mse_loss.py
```diff
@@ -20,29 +20,34 @@ def __init__(self, *args: Any, **kwargs: Any) -> None:
             **kwargs: Keyword arguments passed to torch.nn.MSELoss.
 
         """
+        if "reduction" in kwargs and kwargs["reduction"] != "none":
+        log.warning(
+            "Ignoring reduction='%s'; this loss requires reduction='none' "
+            "for weighted loss computation.", kwargs["reduction"]
+        )
```
marianovitasari20:
```suggestion
          log.warning(
              "Ignoring reduction='%s'; this loss requires reduction='none' "
              "for weighted loss computation.", kwargs["reduction"]
          )
```

icenet_mp/losses/weighted_bce_loss.py
```diff
@@ -20,23 +20,34 @@ def __init__(self, *args: Any, **kwargs: Any) -> None:
...
+        if "reduction" in kwargs and kwargs["reduction"] != "none":
+        log.warning(
+            "Ignoring reduction='%s'; this loss requires reduction='none' "
+            "for weighted loss computation.", kwargs["reduction"]
+         )
```
marianovitasari20:
```suggestion
          log.warning(
              "Ignoring reduction='%s'; this loss requires reduction='none' "
              "for weighted loss computation.", kwargs["reduction"]
           )
```

marianovitasari20:
> > We're overriding it intentionally. The weighted loss requires elementwise outputs (instead of scalar) before applying the weights, so "none" is the only valid value here.
> 
> Do we instead want to raise an exception if the config is providing another value?

done!

jemrobinson (approved):
LGTM

marianovitasari20:
> LGTM

thank you for the reviews @jemrobinson!

PR #307, opened 2026-06-11
Merge touching date ranges in SingleDataset (#279)

### LINKED ISSUE #279: Investigate validation split failure for non-default sample dataset splits
### Problem / Goal

When using sample datasets, changing the validation split date from the default configuration causes the data loading process to fail. This results in a runtime error during dataset slicing/reshaping in the DataLoader, preventing training or evaluation from proceeding.

### What does good look like?

- [ ] Validation split dates work correctly for all supported configurations in sample datasets
- [ ] Data loading completes without errors regardless of the chosen split date
- [ ] If data is not available for a given split configuration, the system should emit a clear warning instead of crashing with a runtime error

### Related context

_No response_

Shaerdan:
Closes #279

  Testing
  - pytest tests/data_loaders/ on an Isambard GH200 node: 48 passed — the new
    regression test, the full single_dataset suite, and the integration layer
    (combined_dataset, common_data_module). No regressions.
  - pre-commit: all hooks pass.
  - mypy: no issues.

  Scope: unit/integration level against the mock-dataset fixture; no end-to-end
  training run on real data (not needed for this change). The change is confined to the date-range-merge
  logic in the data loader.

### DIFF
--- a/icenet_mp/data_loaders/single_dataset.py
+++ b/icenet_mp/data_loaders/single_dataset.py
@@ -1,3 +1,4 @@
+import logging
 from collections.abc import Sequence
 from functools import cached_property
 from pathlib import Path
@@ -12,6 +13,31 @@ from icenet_mp.types import ArrayCHW, ArrayTCHW, DataSpace, Hemisphere
 from icenet_mp.utils import normalise_date
 
 ArrayType = TypeVar("ArrayType", ArrayCHW, ArrayTCHW)
+
+logger = logging.getLogger(__name__)
+
+# helpers: check if any two stacks touch (stack B start is consecutive to stack A end)  --> merge if yes  
+def _ranges_touch(
+    prev: dict[str, str | None],
+    nxt: dict[str, str | None],
+    frequency: np.timedelta64,
+) -> bool:
+    """Whether two ranges touch at the end points.
+
+    Yes when the next range start_date is consecutive to the
+    previous range end_date, ie, no missing day between them (overlap counts
+    too). Open-ended (None) bounds are out of scope, dont merge -- diverse to future fixes/discussions.
+    """
+    if prev["end"] is None or nxt["start"] is None:
+        return False
+    return np.datetime64(nxt["start"]) <= np.datetime64(prev["end"]) + frequency
+
+
+def _later_end(a: dict[str, str | None], b: dict[str, str | None]) -> str | None:
+    """Return the end of the merged span."""
+    if a["end"] is None or b["end"] is None:
+        return None
+    return a["end"] if np.datetime64(a["end"]) >= np.datetime64(b["end"]) else b["end"]
 
 
 class SingleDataset(Dataset):
@@ -73,6 +99,41 @@ class SingleDataset(Dataset):
         return idx2anemoi
 
     @cached_property
+    def _merged_date_ranges(self) -> list[dict[str, str | None]]:
+        """Fuse touching/overlapping ranges into single spans.
+
+        Adjacent ranges touching/overlapping must merge into one
+        single Anemoi subset, or else a run that straddles the range gap/seam is
+        approved against the time step outside the range but fetching is still restricted 
+        to be within the range only, therefore will comes back short, and fails to reshape (issue #279). 
+        Merges are logged to remind the user that their config ranges are touching.
+        """
+        if len(self._date_ranges)  <=  1:
+            return [dict(date_range) for date_range in self._date_ranges]
+        
+        frequency = np.timedelta64(self.load_dataset(self._input_files).frequency)
+        merged: list[dict[str, str | None]] = []
+        for date_range in self._date_ranges:
+            if merged and _ranges_touch(merged[-1], date_range, frequency):
+                prev = merged.pop()
+                fused: dict[str, str | None] = {
+                    "start": prev["start"],
+                    "end": _later_end(prev, date_range),
+                }
+                logger.warning(
+                    "Merged touching date ranges %s and %s into %s for dataset %r "
+                    "Please check cofig file range setting if unintended",
+                    prev,
+                    date_range,
+                    fused,
+                    self._name,
+                )
+                merged.append(fused)
+            else:
+                merged.append(dict(date_range))
+        return merged
+
+    @cached_property
     def dataslices(self) -> list[AnemoiDataset]:
         """Get all slices of contiguous dates from the underlying Anemoi dataset."""
         return [
@@ -82,7 +143,7 @@ class SingleDataset(Dataset):
                 end=date_range["end"],
                 **({"select": self._variables} if self._variables else {}),
             )
-            for date_range in self._date_ranges
+            for date_range in self._merged_date_ranges
         ]
 
     @cached_property
--- a/icenet_mp/data_loaders/single_dataset.py
+++ b/icenet_mp/data_loaders/single_dataset.py
@@ -79,15 +79,14 @@ class SingleDataset(Dataset):
     ) -> list[dict[str, str | None]]:
         """Sort the ranges, then merge overlapping or touching ones into single spans.
 
-        Ranges that overlap or touch must merge into one Anemoi subset, or a run
-        that crosses the seam is approved against a date outside the range while
-        the fetch stays inside it, so the read comes back short and fails to
-        reshape. Two ranges merge when the later one starts on or before the day
-        after the earlier one ends; the merged span runs from the earliest start
-        to the latest end. A "None" bound is open: a "None" start is before any
-        date and a "None" end is after any date, so an open bound always overlaps
-        its neighbour. Each merge is logged so a user notices that their config
-        ranges were altered (e.g. a mistyped year that creates an overlap).
+        Ranges that overlap/touch must merge into one Anemoi subset, or data fetching step comes back
+        short and fails to reshape. Two ranges merge when the later one starts on/before the day
+        after the earlier one finishes; the merged span is the union of all ranges.
+
+        A "None" bound is open, an open-ended previous range always overlap with the later ranges, an
+        open-started later range always overlap with previous ranges.
+
+        Each merge is logged so a user notices that their config ranges were altered and verifies.
         """
         ranges = sorted(
             date_ranges, key=lambda dr: "" if dr["start"] is None else dr["start"]
@@ -103,11 +102,10 @@ class SingleDataset(Dataset):
         ) -> bool:
             """Whether nxt overlaps or is consecutive with prev, so they should merge.
 
-            Ranges are sorted by start, so nxt never begins before prev. They
-            merge when nxt starts on or before the day after prev ends. An open
-            end on prev (None = after any date) covers everything later, and an
-            open start on nxt (None = before any date) sits inside prev, so in
-            either case they always merge.
+            Ranges are sorted by start. They merge when nxt starts on or before the
+            day after prev ends. An open end on prev (None = after any date) covers
+            everything later, and an open start on nxt (None = before any date) sits
+            inside prev, so in either case they always merge.
             """
             if prev["end"] is None or nxt["start"] is None:
                 return True
--- a/icenet_mp/data_loaders/single_dataset.py
+++ b/icenet_mp/data_loaders/single_dataset.py
@@ -77,15 +77,17 @@ class SingleDataset(Dataset):
     def normalise_date_ranges(
         date_ranges: Sequence[dict[str, str | None]],
     ) -> list[dict[str, str | None]]:
-        """Sort the ranges, then merge touching ones into single spans.
+        """Sort the ranges, then merge overlapping or touching ones into single spans.
 
-        Adjacent ranges that touch must merge into one Anemoi subset, or a run
+        Ranges that overlap or touch must merge into one Anemoi subset, or a run
         that crosses the seam is approved against a date outside the range while
         the fetch stays inside it, so the read comes back short and fails to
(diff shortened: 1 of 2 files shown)


icenet_mp/data_loaders/single_dataset.py
```diff
@@ -72,6 +99,41 @@ def _idx2anemoi(self) -> dict[int, tuple[int, int]]:
                     idx2anemoi[idx_global] = (idx_ds, idx_date)
         return idx2anemoi
 
+    @cached_property
```
jemrobinson: This looks good to me, but I think it would be even better to just calculate the merged date ranges once (since overlapping date ranges aren't very useful). 

If you change `def _merged_date_ranges(self) -> list[dict[str, str | None]]:` to `def normalise_date_ranges(self, date_ranges: Sequence[dict[str, str | None]]) -> list[dict[str, str | None]]:` and make the return value `return sorted(merged, key=lambda dr: "" if dr["start"] is None else dr["start"])` then we can simply have `self._date_ranges = self.normalise_date_ranges(data_ranges)` on l.65.

icenet_mp/data_loaders/single_dataset.py
```diff
@@ -13,6 +14,32 @@
 
 ArrayType = TypeVar("ArrayType", ArrayCHW, ArrayTCHW)
 
+logger = logging.getLogger(__name__)
+
+
+# helpers: check if any two stacks touch (stack B start is consecutive to stack A end)  --> merge if yes
```
jemrobinson: Since these helpers are only used in one place, would it be better to have them as nested functions in `_merged_date_ranges`?

icenet_mp/data_loaders/single_dataset.py
```diff
@@ -13,6 +14,32 @@
...
+def _ranges_touch(
+    prev: dict[str, str | None],
+    nxt: dict[str, str | None],
+    frequency: np.timedelta64,
+) -> bool:
+    """Whether two ranges touch at the end points.
+
+    Yes when the next range start_date is consecutive to the
+    previous range end_date, ie, no missing day between them (overlap counts
+    too). Open-ended (None) bounds are out of scope, dont merge -- diverse to future fixes/discussions.
```
jemrobinson: `start=None` means "from the beginning of available data" and `end=None` means "until the end of available data" if that helps with the merging logic.

jemrobinson (changes requested):
A couple of minor suggestions, but this LGTM.

icenet_mp/data_loaders/single_dataset.py
Shaerdan: ah ok that's very helpful, in that case the merge-code doesnt need to worry about them

jemrobinson: Not quite, I think we should use "start=None is before any specified start date" and "end=None is after any specified end date" when calculating overlap.

icenet_mp/data_loaders/single_dataset.py
Shaerdan: good catch; i wasnt sure if i could assume the data is always in daily frequency, so i instead read metadata lazily from the source to get frequency for each source, which is needed to decide if two ranges touch each other; then i needed to have "file list" and "name", ended up having to put the merge method outside of the __init__; maybe this is not needed, i should assume 1 day time step universally and drop the look-up, then put merge method inside __init__ together with sort; 

but before i do this change, can i just confirm with you that: we dont plan to incorporate different frequency of data with this pipeline? e.g., hourly forecast or weekly or sth? if we do, then maybe we could leave these changes as they are, because they could generalize a little bit better

jemrobinson: We might want to incorporate weekly or monthly in future. However, this will likely be run on top of daily datasets (e.g. only selecting every 7th entry from a dataset that contains daily data).

Shaerdan: ok thx, I will fix frequency to daily and simplify _merged_date_ranges and make it part of the __init__

Shaerdan: all suggested changes implemented and done in 49de94c6

icenet_mp/data_loaders/single_dataset.py
```diff
@@ -72,6 +73,59 @@ def _idx2anemoi(self) -> dict[int, tuple[int, int]]:
...
+
+        #  Assume data always has a daily freuquency; for future updates: read from source metadata
+        frequency = np.timedelta64(1, "D")
+
+        def _ranges_touch(
+            prev: dict[str, str | None], nxt: dict[str, str | None]
+        ) -> bool:
+            """Check whether nxt starts within one time-step of prev's end."""
+            if prev["end"] is None or nxt["start"] is None:
+                return False
```
jemrobinson: Should this be?

```suggestion
            if prev["end"] is None or nxt["start"] is None:
                return True
```

since a range with "end=None" will overlap any range starting after it and a range with "start=None" will overlap any range starting before it?

jemrobinson (changes requested):
Can you confirm the logic in the date-touching is correct? Maybe let's add a test for start=None and one for end=None?

icenet_mp/data_loaders/single_dataset.py
Shaerdan: oh yes that is a legacy left-over I missed, before i knew the meaning of None in str and end, I tried to leave out any case of "None" appearing in the joint of any two range. obviously not needed anymore and i will get rid of that; thx

icenet_mp/data_loaders/single_dataset.py
```diff
@@ -99,12 +99,17 @@ def normalise_date_ranges(
...
+            left untouched.
+            """
             if prev["end"] is None or nxt["start"] is None:
                 return False
-            return bool(
-                np.datetime64(nxt["start"]) <= np.datetime64(prev["end"]) + frequency
-            )
+            end = np.datetime64(prev["end"])
+            start = np.datetime64(nxt["start"])
+            return bool(end <= start) and bool(start <= end + frequency)
```
Shaerdan, line 121: the previous touch detection logic could silently include a scenario of overlapping ranges and then crop out valid data mistakenly, if overlapping ranges was given by the user's config -- which we assumed wont happen anyway but the code logic shouldnt silently mishandle them in case they do happen;  the correctly logic will just skip any overlapping case, leave them untouched. we only merge when two ranges touch each other consecutively at the joint.

tests/data_loaders/test_single_dataset.py
```diff
@@ -432,3 +432,62 @@ def test_subset_preserves_normalise_flag(self, mock_dataset: Path) -> None:
             )
             subset = dataset.subset(variables=["ice_conc"])
             assert subset._normalise is flag
```
Shaerdan, line 435: a list of unit tests below were conducted and all passed

tests/data_loaders/test_single_dataset.py
```diff
@@ -412,3 +432,62 @@ def test_subset_preserves_normalise_flag(self, mock_dataset: Path) -> None:
...
+        ]
+        assert SingleDataset.normalise_date_ranges(ranges) == ranges
+
+    def test_normalise_date_ranges_start_none_left_alone(self) -> None:
+        """Two open starts (None = first date) overlap, so they are left alone."""
+        ranges: list[dict[str, str | None]] = [
+            {"start": None, "end": "2020-03-31"},
+            {"start": None, "end": "2020-06-30"},
+        ]
+        assert SingleDataset.normalise_date_ranges(ranges) == ranges
```
jemrobinson: I think that in this case we should combine the ranges into a single range `{"start": None, "end": "2020-06-30"}`. Do you think this is a bad idea?

tests/data_loaders/test_single_dataset.py
```diff
@@ -412,3 +432,62 @@ def test_subset_preserves_normalise_flag(self, mock_dataset: Path) -> None:
...
+
+    def test_normalise_date_ranges_end_none_left_alone(self) -> None:
+        """An open end (None = last date) overlaps a later range, so it is left alone."""
+        ranges: list[dict[str, str | None]] = [
+            {"start": "2020-01-01", "end": None},
+            {"start": "2020-06-01", "end": "2020-09-30"},
+        ]
+        assert SingleDataset.normalise_date_ranges(ranges) == ranges
```
jemrobinson: I think that in this case we should combine the ranges into a single range `{"start": "2020-01-01", "end": None}`. Do you think this is a bad idea?

tests/data_loaders/test_single_dataset.py
```diff
@@ -412,3 +432,62 @@ def test_subset_preserves_normalise_flag(self, mock_dataset: Path) -> None:
...
+            {"start": "2020-01-01", "end": "2020-12-31"},
+            {"start": "2020-03-01", "end": "2020-06-30"},
+        ]
+        assert SingleDataset.normalise_date_ranges(ranges) == ranges
+
+    def test_normalise_date_ranges_end_none_left_alone(self) -> None:
+        """An open end (None = last date) overlaps a later range, so it is left alone."""
+        ranges: list[dict[str, str | None]] = [
+            {"start": "2020-01-01", "end": None},
+            {"start": "2020-06-01", "end": "2020-09-30"},
```
jemrobinson: Perhaps it's worth considering:

```
{"start": "2020-06-01", "end": None},
{"start": "2020-01-01", "end": "2020-09-30"},
```

which I think should resolve to `{"start": "2020-01-01", "end": None}`

jemrobinson (changes requested):
Some thoughts on how a couple of these test scenarios should work. Thoughts @IFenton @marianovitasari20 ?

tests/data_loaders/test_single_dataset.py
marianovitasari20: I think I agree with handling any overlapping or consecutive dates on our side by merging them and taking the earliest start date and latest end date.

For example, if one period is from January to March 2025 and another is from February to May 2025, we can merge them into a single period with a start date of January 2025 and an end date of May 2025.

Shaerdan: i also lean on the strategy of making it just fully fool-proof, since we are already covering the touching range cases might as well cover all other possible cases that should be merged. also i should mention: i forgot to remove the dummy guard in the touch detector again...  this part:       
      l. 108 -            if prev["end"] is None or nxt["start"] is None:
      l. 109 -                return False

I will clean them in the next update once we decided the strategy to go

IFenton:
> Some thoughts on how a couple of these test scenarios should work. Thoughts @IFenton @marianovitasari20 ?

Those seem like reasonable solutions. However it occurs to me that we should probably alert the user to fact we are doing a merge. It's not very often that these dates will be changed, and I would think there is a good chance that overlaps occurs by accident when the user actually meant to do something else (e.g. they put in the wrong year). Would it be possible to print out the ranges that are actually being used / alert the user to the fact that we've updated their ranges so they can check that it was what they intended?

Also, do we have checks for whether any of the train / test / validate periods overlap?

marianovitasari20:
> > Some thoughts on how a couple of these test scenarios should work. Thoughts @IFenton @marianovitasari20 ?
> 
> Those seem like reasonable solutions. However it occurs to me that we should probably alert the user to fact we are doing a merge. It's not very often that these dates will be changed, and I would think there is a good chance that overlaps occurs by accident when the user actually meant to do something else (e.g. they put in the wrong year). Would it be possible to print out the ranges that are actually being used / alert the user to the fact that we've updated their ranges so they can check that it was what they intended?
> 
> Also, do we have checks for whether any of the train / test / validate periods overlap?

I think that's a good point. I agree with adding a warning showing what was input and what it changed to, to make the user aware. And as @IFenton said, in case there are overlapping training/val/test periods, we should also show a warning message for them to check if their input is correct, and warn about overlaps between training and val, etc.

icenet_mp/data_loaders/single_dataset.py
jemrobinson: I'm not sure what this means. Can you add a test case that covers the scenario you have in mind?

Shaerdan: that was from the previous version where we probably said "fixing overlapping ranges" might not be very useful, which i took as users dont normally write overlapping ranges; i havent pushed the latest commit yet, will do now and it should have all cases covered and tested  (touching and overlapping both included) for you to review :)

tests/data_loaders/test_single_dataset.py
```diff
@@ -412,3 +433,116 @@ def test_subset_preserves_normalise_flag(self, mock_dataset: Path) -> None:
...
+        ) == [{"start": None, "end": "2020-06-30"}]
+
+    def test_normalise_date_ranges_open_outer_ends_merge(self) -> None:
+        """Open-beginning + open-ending halves that meet merge to the whole range."""
+        assert SingleDataset.normalise_date_ranges(
+            [
+                {"start": None, "end": "2020-06-30"},
+                {"start": "2020-07-01", "end": None},
+            ]
+        ) == [{"start": None, "end": None}]
```
jemrobinson, line 515: Should we have a test for non-touching overlaps

```
{"start": None, "end": "2020-07-31"},
{"start": "2020-06-01", "end": None},
```

or is this adequately covered by `test_normalise_date_ranges_partial_overlap_merge`?

jemrobinson (approved):
LGTM. One thought about a possible additional test, but I'm not sure it's needed. Feel free to merge when you're ready @Shaerdan .

PR #313, opened 2026-06-19
Autoregressive ddpm

marianovitasari20:
Adds an autoregressive sampling mode to the DDPM model, resolving issue #237. 

Previously the model only supported parallel (non-autoregressive) sampling, where all forecast steps were generated simultaneously as channels in a single diffusion process. This PR introduces a sequential alternative where each forecast step is produced by its own independent reverse diffusion process, conditioned on previously predicted outputs.

For example, with `n_history_steps` = 3:
**OSISAF (SIC) conditioning:**
[t-3, t-2, t-1] → predict t        (t   is observed)
[t-2, t-1,  t ] → predict t+1      (t   is model-predicted)
[t-1,  t , t+1] → predict t+2      (t, t+1 are model-predicted)
...
**ERA5 (atmospheric) conditioning:**
era5_forecast available     → use the ERA5 forecast frame for that step
era5_forecast not available → repeat the last observed ERA5 frame

W&B: https://wandb.ai/turing-seaice/train/runs/cgtit4yp
image

https://github.com/user-attachments/assets/ffa1d987-fb9e-449d-aef7-50d06037866c

https://github.com/user-attachments/assets/38cd45b7-76a3-4bc3-a9a2-6794e678d321

https://github.com/user-attachments/assets/d722b5b8-f41f-4698-b542-e0554e8933c5


marianovitasari20:
This PR is ready for review!
I've added more documentation than usual in the code. I normally include detailed documentation, but since questions around DDPM tend to come up in almost every meeting, I figured it's better to have longer and much more explanation in each function, especially for people who are less familiar with the codebase, so it can help clear up the confusion. Hope it helps!
If you have any questions, feel free to ask. I'm happy to answer during today's meeting!

icenet_mp/models/ddpm.py
jemrobinson, line 155: These are duplicate definitions (see lines 106-111)

icenet_mp/models/ddpm.py
```diff
@@ -383,9 +543,8 @@ def test_step(
             - loss: test loss value
 
         """
-        x = self.prepare_inputs(batch)  # [B, C_cond, H, W]
         y = batch["target"]  # [B, T, 1, H, W]
```
jemrobinson: Why is validation not also autoregressive?

icenet_mp/models/ddpm.py
```diff
@@ -155,14 +171,84 @@ def forward(self, *args: Any, **kwargs: Any) -> NoReturn:
         msg = "This model uses `training_step`, `validation_step`, and `test_step` instead of `forward()`"
         raise NotImplementedError(msg)
 
-    def sample(self, x: TensorNCHW) -> TensorNCHW:
-        """Perform reverse diffusion sampling starting from noise.
+    def sample(
+        self,
+        batch: dict[str, TensorNTCHW],
+        sample_weight: torch.Tensor | None = None,  # noqa: ARG002
```
jemrobinson: `sample_weight` isn't used. Can we drop it from the function signature?

icenet_mp/models/ddpm.py
```diff
@@ -383,9 +543,8 @@ def test_step(
             - loss: test loss value
 
         """
-        x = self.prepare_inputs(batch)  # [B, C_cond, H, W]
         y = batch["target"]  # [B, T, 1, H, W]
-        y_hat = self.sample(x).unsqueeze(2)  # [B, C_cond, 1, H, W]
+        y_hat = self.sample(batch).unsqueeze(2)  # [B, C_cond, 1, H, W]
```
jemrobinson: Is this really `C_cond`? I think it should be `n_forecast_steps * base_output_channels`.

icenet_mp/models/ddpm.py
```diff
@@ -155,14 +171,84 @@ def forward(self, *args: Any, **kwargs: Any) -> NoReturn:
         msg = "This model uses `training_step`, `validation_step`, and `test_step` instead of `forward()`"
         raise NotImplementedError(msg)
 
-    def sample(self, x: TensorNCHW) -> TensorNCHW:
-        """Perform reverse diffusion sampling starting from noise.
+    def sample(
+        self,
+        batch: dict[str, TensorNTCHW],
+        sample_weight: torch.Tensor | None = None,  # noqa: ARG002
+    ) -> torch.Tensor:
```
jemrobinson:
```suggestion
    ) -> TensorNCHW:
```

icenet_mp/models/ddpm.py
```diff
@@ -188,6 +274,80 @@ def sample(self, x: TensorNCHW) -> TensorNCHW:
...
+            all_predictions.append(y_step)
+
+            # Slide OSISAF window: append prediction as new frame
+            osisaf = torch.cat([osisaf, y_step.unsqueeze(1)], dim=1)
+
+            # Slide ERA5 window: use forecast if available, else repeat last frame
+            if era5_forecast is not None:
+                next_era5 = era5_forecast[:, step : step + 1]
+            else:
+                next_era5 = era5[:, -1:]
```
jemrobinson, line 380: We should warn when this happens, at least the first time.

icenet_mp/models/ddpm.py
```diff
@@ -270,7 +430,10 @@ def training_step(
         x = self.prepare_inputs(batch)  # [B, C_cond, H, W]
 
         # Extract target
-        y = batch["target"].squeeze(2)  # B, T, H, W
+        if self.use_autoregressive:
+            y = batch["target"][:, 0]  # [B, 1, H, W] — one step at a time
```
jemrobinson: Does this mean that we only use the first timestep to calculate the loss in training?

jemrobinson (changes requested):
LGTM overall, but I have some questions.

icenet_mp/models/ddpm.py
marianovitasari20: Fair point. Yes, it's not being used yet. I put it there in case we want to use sample weighting in the future. Happy to remove if you'd prefer.

icenet_mp/models/ddpm.py
marianovitasari20: Sorry, missed updating this comment. You're right, it should be `n_forecast_steps * base_output_channels ` (which  equals to `T`, since `base_output_channels` is 1). Fixing it.

marianovitasari20:
```suggestion
        y_hat = self.sample(batch).unsqueeze(2)  # [B, T, 1, H, W]
```

icenet_mp/models/ddpm.py
marianovitasari20: done in https://github.com/alan-turing-institute/icenet-mp/pull/313/changes/4c94e510bb7a7d652c9092f2c0a29f204a4a9747

icenet_mp/models/ddpm.py
marianovitasari20: It's actually autoregressive. If you look at sample(), it branches on self.use_autoregressive and routes to _sample_autoregressive when it's set, so test/validation do get autoregressive sampling.

icenet_mp/models/ddpm.py
marianovitasari20: Good catch, thanks James. You're totally right! I accidentally pasted it twice when moving this block.

marianovitasari20: done in https://github.com/alan-turing-institute/icenet-mp/pull/313/changes/b0813d95b8568c876bf00169f11830d56fb7bacf

icenet_mp/models/ddpm.py
marianovitasari20: Since we are doing autoregressive, we only care about predicting T_0 using historical data/input during training. For the remaining time steps, we use a rolling window during inference, that is the logic behind my implementation. I'm currently double-checking again and debugging the code to ensure this logic is correct.

icenet_mp/models/ddpm.py
```diff
@@ -188,6 +270,89 @@ def sample(self, x: TensorNCHW) -> TensorNCHW:
 
         return torch.clamp(y, 0, 1)
```
jemrobinson:
```suggestion
        return self.mask(self.restrict(y))
```

icenet_mp/models/ddpm.py
```diff
@@ -188,6 +270,89 @@ def sample(self, x: TensorNCHW) -> TensorNCHW:
...
+            y = torch.randn((B, self.base_output_channels, H, W), device=self.device)
+
+            # Diffusion reverse process
+            for t in reversed(range(self.timesteps)):
+                t_batch = torch.full((B,), t, dtype=torch.long, device=self.device)
+                pred_v: torch.Tensor = self.model(y, t_batch, x)
+                y = self.diffusion.p_sample(y, t_batch, pred_v)
+
+            # Clamp and store prediction
+            y_step = torch.clamp(y, 0, 1)
```
jemrobinson:
```suggestion
            y_step = self.mask(self.restrict(y))
```

jemrobinson:
@marianovitasari20 : can you merge main into this branch and check everything works as expected? It would also be good to see a comparison of the autoregressive method against the old method (this can just be on the sample dataset if that's easier).

jemrobinson:
@marianovitasari20 : after you've merged main into this branch, can you check that the logic deals with the SSMIS dataset correctly? It has 6 channels, so the NTCHW shape is e.g. [2, 3, 6, 432, 432] and you can't assume that `squeeze(2)` will just pull out the target channel.

marianovitasari20:
Still in progress. I'm still double-checking it due to the recent changes in `main`.

marianovitasari20:
I haven't committed my last change because I haven't managed to debug it properly yet. It's still only in my local branch. I'm having problems with GPU resources, Isambard has been at full capacity recently. I'm trying to run it locally, but it's still downloading the data.

jemrobinson:
> I'm trying to run it locally, but it's still downloading the data.

Have you tried downloading the datasets from Isambard? They should also be available there.

marianovitasari20:
> > I'm trying to run it locally, but it's still downloading the data.
> 
> Have you tried downloading the datasets from Isambard? They should also be available there.

I decided to just run it on one month of data for debugging purposes, and it works now. I'll do the full run later once we have enough compute resources. Thanks for the suggestion!

marianovitasari20:
This is the W&B link for the autoregressive run after merging with `main` (including the shape changes based on SSMIS):

https://wandb.ai/turing-seaice/train/runs/p0lfm0vb

image

For the non-autoregressive run, the job is still queueing.

jemrobinson:
Some notes on the above run (before we lose Isambard access):

```yaml
datasets:
  samp-floatsouth-argo-25p0km-2017-2019-24h-v2
  samp-sicsouth-ssmis-25p0km-2017-2019-24h-v2
  samp-weathersouth-era5-25p0km-2017-2019-24h-v3
split:
  train:
  - start: '2017-01-01'
    end: '2017-12-31'
  - start: '2018-02-01'
    end: '2018-12-31'
  validate:
  - start: '2018-01-01'
    end: '2018-01-31'
loss:
  _target_: torch.nn.HuberLoss
  delta: 0.5
model:
  _target_: icenet_mp.models.ddpm.DDPM
  name: ddpm
  dropout_rate: 0.05
  mask_type: active
  restrict_range: clamp
  start_out_channels: 64
  timesteps: 150
predict:
  n_forecast_steps: 2
  n_history_steps: 3
random:
  seed: 555
  fully_deterministic: false
```

icenet_mp/models/ddpm.py
```diff
@@ -205,21 +284,104 @@ def sample(self, x: TensorNCHW) -> TensorNCHW:
...
+        corresponding ERA5 forecast frame (if provided) or a repetition of the last
+        observed ERA5 frame is appended to the ERA5 history.
+
+        Args:
+            batch (dict[str, TensorNTCHW]):
+                Dictionary containing:
+                    - self.osisaf_key: OSISAF SIC history tensor of shape
+                    [B, T, 1, H, W].
+                    - era5: ERA5 input tensor of shape [B, T, C, H2, W2].
+                    - era5_forecast (optional): ERA5 forecast tensor of shape
```
jemrobinson, line 317: At the moment, do we ever provide `era5_forecast`? Is it possible for the user to do so?

icenet_mp/models/ddpm.py
```diff
@@ -295,29 +457,38 @@ def training_step(
...
 
         """
         # Prepare input tensor by combining osisaf-south and era5
         x = self.prepare_inputs(batch)  # [B, C_cond, H, W]
 
         # Extract target
-        y = batch["target"].squeeze(2)  # B, T, H, W
+        if self.use_autoregressive:
+            y = batch["target"][
+                :, 0
```
jemrobinson:
```suggestion
                :, 0, :, :, :
```

this has the same meaning but is maybe clearer about which channel you are selecting from?

jemrobinson:
I wanted to test that both paths would run, so I ran a local 2-epoch test with the following results:

## autoregressive=True
- [comfy-dust-1394](https://wandb.ai/turing-seaice/train/runs/j59ja718)
- local running on macOS: `0.73it/s`
media_images_output_static_2018-01-16-sea-ice-concentration_16_40ce727cbe16c060005c

## autoregressive=False
- [happy-sun-1395](https://wandb.ai/turing-seaice/train/runs/vhwoawpw)
- local running on macOS: `0.68it/s`
media_images_output_static_2018-01-16-sea-ice-concentration_16_b3d5e3272c4253c06f65

icenet_mp/models/ddpm.py
```diff
@@ -205,21 +284,104 @@ def sample(self, x: TensorNCHW) -> TensorNCHW:
...
+            new_frame = osisaf[:, -1:].clone()  # [B, 1, C_in, H, W]
+            new_frame[:, :, : self.base_output_channels] = y_step.unsqueeze(1)
+            osisaf = torch.cat([osisaf, new_frame], dim=1)
+
+            # Slide ERA5 window: use forecast if available, else repeat last frame
+            if era5_forecast is not None:
+                next_era5 = era5_forecast[:, step : step + 1]
+            else:
+                if not self._warned_missing_era5_forecast:
+                    warnings.warn(
```
jemrobinson: I think `logging.warning` might be better here, as this message is more like "here's a thing you should know but it's not clear how to fix it" than "you've configured this incorrectly - please fix" (which would be a good use-case for `warn.warning`).

jemrobinson (changes requested):
LGTM - a couple of minor comments.

icenet_mp/models/ddpm.py
marianovitasari20: This is currently just a placeholder for a future feature (era5_forecast or any other data source used as input for the next autoregressive step). At the moment, I use repeated history, and only SIC is autoregressive. The placeholder is there so that, later when we want to feed actual ERA5 forecast data (rather than historical reanalysis) into the model during autoregressive rollout, we only need to add the relevant config/dataset instead of restructuring the code path. Happy to add a comment if it is confusing.
and yes, it's possible to do so if we have the data.

icenet_mp/models/ddpm.py
marianovitasari20: done in https://github.com/alan-turing-institute/icenet-mp/pull/313/changes/fbb93d2f24ff57772e6e423191a2e9e38eec7376

marianovitasari20:
**W&B links** after merging with `main` (including the shape changes based on SSMIS):

> **Autoregressive**
https://wandb.ai/turing-seaice/train/runs/p0lfm0vb
media_images_output_static_2018-01-16-sea-ice-concentration_1738_6d929553a2e08fe82df1


> **Non-autoregressive**
https://wandb.ai/turing-seaice/train/runs/kleyg5an
media_images_output_static_2018-01-16-sea-ice-concentration_1738_1fd1b9340c7b87d59d5f

marianovitasari20:
image

jemrobinson (approved):
LGTM

marianovitasari20:
> LGTM

Many thanks @jemrobinson for the reviews! :)

PR #354, opened 2026-07-03
Add pre-commit to dependabot

jemrobinson:
I noticed our precommit actions are a bit out of date so I've added them to Dependabot.


IFenton (approved):
LGTM

Issue #272, opened 2026-05-15
Update the model to incorporate all history days into each prediction


IFenton:
### Problem / Goal

Currently the model uses the first of the history days to predict the first of the forecast days, the second of the history days to predict the second forecast day etc. So:

```
[H:-3] -> [F:1]
[H:-2] -> [F:2]
[H:-1] -> [F:3]
[F:1] -> [F:4]
etc.
```

Going forward it would be good to use all of the history days to predict the first forecast day, and then all but one of the history days and the first forecast day to predict the second. So:

```
[H:-3, H:-2, H:-1] -> [F:1]
[H:-2, H:-1, F:1] -> [F:2]
[H:-1, F:1, F:2] -> [F:3]
etc.
```

### What does good look like?

- [ ] Implement the new structure in the pipeline

IFenton:
Is this correct @jemrobinson?

jemrobinson:
Note that the current setup (using one timestep to predict one timestep) has duplication across batches

## Current setup 

batch 1:
```
t0 -> t3*
t1 -> t4*
t2 -> t5*
t3* -> t6*
```

batch 2:
```
t1 -> t4*
t2 -> t5*
t3 -> t6*
t4* -> t7*
```

Note that e.g. `t4*` has identical inputs in both batches

## Multi-timestep proposal

batch 1:
```
cat[t0, t1, t2] -> t3*
cat[t1, t2, t3*] -> t4*
cat[t2, t3*, t4*] -> t5*
cat[t3*, t4*, t5*] -> t6*
```

batch 2:
```
cat[t1, t2, t3] -> t4'
cat[t2, t3, t4'] -> t5'
cat[t3, t4', t5'] -> t6'
cat[t4', t5', t6'] -> t7'
```

Note that e.g. `t4*` has different inputs from `t4'`

louisavz:
@jemrobinson asked me to comment here from our discussion with @ots22. Not sure how much more I can add since James already summarised the conclusion nicely. 

It's not ideal doing what we currently are doing via the "leap frog" approach. The multi-timestep to make one prediction makes more sense and it's what I thought we were doing.

@IFenton suggested change makes more sense, with some adjustment to the notation to match James'.

```
[H:-2, H:-1, H:0] -> [F:1]
[H:-1, H:0, F:1] -> [F:2]
[H:0, F:1, F:2] -> [F:3]
etc.
```

jemrobinson:
[...]

PR #363, opened 2026-07-07
Fold history dimension into channels for processing

AoifeHughes:
- Addresses #272 

rollout now keeps a window of the n_history_steps most recent timesteps, cats them on the channel dim, feeds that to forward, then slides the window (drop oldest, append prediction)


icenet_mp/models/processors/base_processor.py
```diff
@@ -72,15 +82,17 @@ def rollout(self, x: TensorNTCHW, y: TensorNTCHW | None = None) -> ProcessorOutp
...
-        # Rollout the model over the input slices, producing an output for each one.
-        # Also append the predictions to the list of input slices, so that we can still
-        # predict when n_forecast_steps > n_history_steps.
+        # Slide the window forward, predicting one timestep at a time from the whole
+        # window and then dropping the oldest timestep to make room for the prediction.
         outputs: list[TensorNCHW] = []
         for _ in range(self.n_forecast_steps):
-            outputs.append(self(nchw_slices.pop(0)))
-            nchw_slices.append(outputs[-1])
+            next_step = self(cat(window, dim=1))
```
jemrobinson: Quick thought here: what actually breaks if we pass NTCHW tensors to the processor instead of NCHW ones? Should be processor-dependent, but it might be that some of them can handle 5D tensors without (much) modification.

icenet_mp/models/processors/base_processor.py
```diff
@@ -72,15 +82,17 @@ def rollout(self, x: TensorNTCHW, y: TensorNTCHW | None = None) -> ProcessorOutp
...
+            outputs.append(next_step)
```
jemrobinson, line 97: Related to the above: if we go down this route, is it better to collapse the T dimension into N or C? Can we justify the choice?

icenet_mp/models/processors/null.py
```diff
@@ -28,13 +28,17 @@ def __init__(self, **kwargs: Any) -> None:
...
         Args:
-            x: TensorNCHW with (batch_size, n_latent_channels_total, latent_height, latent_width)
+            x: TensorNCHW with (batch_size, n_latent_channels_total * n_history_steps, latent_height, latent_width)
 
         Returns:
             TensorNCHW with (batch_size, n_latent_channels_total, latent_height, latent_width)
 
         """
-        return self.model(x)
+        return self.model(x[:, -self.data_space.channels :, :, :])
```
jemrobinson: Would this be clearer as `self.model(x)[:, -self.data_space.channels :, :, :]`? I know they're equivalent but conceptually one is "slice the input" and the other is "slice the output".

tests/models/test_processors.py
```diff
@@ -45,6 +45,78 @@ def test_rollout(
             )
 
 
+class RecordingProcessor(BaseProcessor):
```
jemrobinson, line 48: I love this. Very nice way to test the window behaviour.

icenet_mp/models/processors/base_processor.py
AoifeHughes: Wouldn't we deliberately chose C over N?

Conv2d/patch-embedding mix across channels but treat batch elements independently, so collapsing into N would give 3 independent single-frame predictions again, which isn't that the exact bug we're fixing?

jemrobinson: That's a good argument. From what I remember [SimVP](https://arxiv.org/abs/2206.05099) collapse into N, but I might be misremembering (and I can't remember whether they discuss it in the paper).

AoifeHughes: From what I could tell SimVP uses N-collapse only for the spatial encoder/decoder, where no temporal mixing is needed (each frame is processed independently on purpose). The part of SimVP responsible for actual temporal modeling uses C-collapse (section 3 of that linked paper)

icenet_mp/models/processors/null.py
AoifeHughes: Yes! Good spot, thank you. This was fixed here: https://github.com/alan-turing-institute/icenet-mp/pull/363/changes/6253b890a72fc65643308e148027ae4ff98e3d30

icenet_mp/models/processors/base_processor.py
AoifeHughes: Also, do you feel that https://github.com/alan-turing-institute/icenet-mp/pull/363#discussion_r3536776205 needs resolved separately - I tried to only reply here to save 2x related conversations.

icenet_mp/models/processors/base_processor.py
jemrobinson: > Also, do you feel that https://github.com/alan-turing-institute/icenet-mp/pull/363#discussion_r3536776205 needs resolved separately - I tried to only reply here to save 2x related conversations.

I think the answer to my question is that `Conv2D` (among others) will break, as it expects NCHW input.

jemrobinson:
Overall, this looks like a sensible, minimal change to support history days. Can we run a couple of comparisons:

- collapsing into N vs. collapsing into C
- whichever of the above is better vs. baseline (i.e. the current `main`)

You might need to merge `main` into this branch to get sensible comparisons. I suggest the `cnn_vit_cnn` model if you can wait until #324 is merged.

AoifeHughes:
> * collapsing into N vs. collapsing into C
> 
> * whichever of the above is better vs. baseline (i.e. the current `main`)

Maybe I'm understanding wrong, but isn't this less about adding this 'to support history days' and more about fixing a genuine bug? I'm asking not out of nitpicking, but to say that even if this was worse than the current main it would still be needed as the current implementation isn't intentional? If the current setup was still producing better results then that's a genuine interest point and should be studied and explored separately. 

I've no strong feelings on this, just thought I'd flag. 

I can run a comparison either way, I did for my non-commited testing so will set something more robust up and share results.

icenet_mp/models/processors/base_processor.py
jemrobinson: Confirmed: `RuntimeError: Expected 3D (unbatched) or 4D (batched) input to conv2d, but got input of size: [2, 3, 288, 144, 144]`. It would be possible to write a UNet that worked with 5D tensors (e.g. using Conv3D) but this isn't a trivial change.

jemrobinson:
> Maybe I'm understanding wrong, but isn't this less about adding this 'to support history days' and more about fixing a genuine bug? I'm asking not out of nitpicking, but to say that even if this was worse than the current main it would still be needed as the current implementation isn't intentional? If the current setup was still producing better results then that's a genuine interest point and should be studied and explored separately.

I agree that we'd probably consider merging this even if it doesn't outperform the current, incorrect implementation. However, it would be good to see some evidence and also, if it *is* worse, we'd like to understand why and possibly consider other options.

> I can run a comparison either way, I did for my non-commited testing so will set something more robust up and share results.

Thanks :)

AoifeHughes:
Only ran for a few epochs so I could compare all, but I think the trends are at least useful enough? I'll run for more epochs if it would be useful 

output_fifo
output-test-set

icenet_mp/models/processors/base_processor.py
```diff
@@ -72,15 +82,17 @@ def rollout(self, x: TensorNTCHW, y: TensorNTCHW | None = None) -> ProcessorOutp
                 loss: an optional processor-specific loss
 
         """
-        # Cut the NTCHW input into NCHW slices
-        nchw_slices = [x[:, idx_t, :, :, :] for idx_t in range(self.n_history_steps)]
+        # The current window of n_history_steps timesteps, oldest to newest
+        window: list[TensorNCHW] = [
```
jemrobinson: Is this better as `window: deque[TensorNCHW] = [` since we want to both add to the right and remove from the left?

If so, then `window = [*window[1:], next_step]` could be

```python
window.popleft()
window.append(next_step)
```

which would avoid copy-in-place.

AoifeHughes: added in last commit

AoifeHughes:
output-extended
Trained some more epochs - C-collapse still fractionally better than N. 
As I'd mentioned earlier I think having this as a default would be fine, and overriding the base-processor to test N collapse could be done easily if people wanted to explore it. 

I've more models still running with more data and longer ranges to match real specs. But if someone is happy to approve and merge, I'd appreciate it.

icenet_mp/models/processors/unet.py
```diff
@@ -50,8 +50,13 @@ def __init__(
         channels = [start_out_channels * 2**exponent for exponent in range(4)]
 
         # Encoder
+        # The input is a window of n_history_steps timesteps concatenated along
+        # channels (see BaseProcessor.rollout), so the first layer must accept
+        # n_history_steps times as many channels as a single timestep has.
         self.conv1 = CommonConvBlock(
-            self.data_space.channels, channels[0], kernel_size=kernel_size
+            self.data_space.channels * self.n_history_steps,
```
louisavz, line 57: Can we move this change outside of the conv1 layer so we are not changing the model architecture?

icenet_mp/models/processors/vit.py
```diff
@@ -45,8 +45,14 @@ def __init__(
         self.patch_size = patch_size
         self.out_channels = self.data_space.channels
 
+        # The input is a window of n_history_steps timesteps concatenated along
+        # channels (see BaseProcessor.rollout), so patch embedding must accept
+        # n_history_steps times as many channels as a single timestep has.
         self.patch_embed = PatchEmbedding(
-            self.data_space.channels, patch_size, emb_dim, self.img_size
+            self.data_space.channels * self.n_history_steps,
```
louisavz, line 52: Likewise, can we move this change outside of the patch_embed layer so we are not changing the model architecture? I believe this would break Erin's checkpoints. I suppose my question is, is it possible or does it make sense to have this fix outside of the models?

AoifeHughes: Wouldn't having the release, like we did today be the proper way to do things if someone wanted to reload checkpoints / support inter-commit checkpoint loading? 

I'm hesitant on both comments simply because it adds complexity and would lead to a bottleneck in having to do the compression prior to conv1. Supporting backwards compatibility with training checkpoints in a rolling library seems like a headache and I'd really only ever re-run with checkpoints associated with the repo at the commit hash for which it was trained.

IFenton: In an aside on backward compatability, even without this change Erin's models won't run on the codebase as it currently is. Any time we change one of the arguments in the configs, as James' PR did earlier today it stops old checkpoints running on the new code.

louisavz: I'm wondering the same. I think some changes are worth making to break backward compatibility, I'm debating if this one is it. It feels cleaner (but more complex, requiring another layer or outside of the models) to not repeat this compression across all models. However, there will likely be models in the future that handles this differently than our existing models, so we don't want to restrict new models to work with collapsed dimension either. I think this is why I'm asking if it makes sense, and I realise I might just talked myself out of it :).

louisavz: Just saw Isabel's comment. How is it changing one argument in the config won't allow Erin to run her checkpoint? Did it change the architecture?

louisavz: I should phrase my question better, and look at the argument being changed in that PR. Changing an argument that affects the model in the processor architecture would not allow us to load the checkpoints, but if it doesn't change the architecture, then it shouldn't break it. Let me review how the checkpoints are being saved next.

louisavz: Okay, I see while James' PR on config didn't make changes that would affect the ViT model, the checkpoints are saved for the pull pipeline rather than the model itself. I think that's where the multi-stage training comes in, introducing some isolation in components. Regarding where to perform the dimension reduction and while I'm leaning toward one direction, there are definitely pros as is. I'll go ahead and resolve this comment and the last. Many thanks.

jemrobinson: I think it's fine to have previous checkpoints tied to particular releases of the code - it gives you a fixed reference point which we don't otherwise have. I have training jobs running for `2026.07` at the moment for our baseline models, which are similar to but not exactly the same as Erin's ones.

jemrobinson:
> more epochs - C-collapse still fractionally better than N. As I'd mentioned earlier I think having this as a default would be fine, and overriding the base-processor to test N collapse could be done easily if people wanted to explore it. 

I'd prefer to leave this as *not* configurable. If someone wants to go back and try N-collapse, it's such a trivial change (1-character!) that they can do so at that point.

jemrobinson (approved):
LGTM. @AoifeHughes : merge whenever you're happy.

PR #371, opened 2026-07-13
Add synthetic data testing framework

### LINKED ISSUE #177: Add continuous Integration tests for the pipeline
### Problem / Goal

Adding integration and continuous integration tests for Erin's model and Maria's model to help with development. Particularly with merging pipelines into the main pipeline, we want to check nothing has changed in the output of their models. 

### What does good look like?

- [ ] Integration tests written for either a subset of the data or dummy data
- [ ] Continuous integration in GitHub actions

### Related context

_No response_

### LINKED ISSUE #367: Add synthetic model test framework for rapid pipeline validation and testing of model's ability to learn
As discussed during co-working session on 9/7/2026. 

I think it would be useful to have some semi-automated testing and validation of model pipelines that use very simple synthetic data. My current thoughts are: 

1. Generate a binary image with a circle drawn in 1's
2. Perform some basic matrix shifting to simulate 'moving' of ice and create a set of test/train/val images
3. Use existing configs to pass this simulated data 
4. Verify outputs are as expected.

If a model couldn't learn a simple linear movement then it would flag and save running larger costly training. Having this simple and efficient enough would be good for CI reasons too to ensure we don't make breaking changes to model logic.

AoifeHughes:
## Summary

  Synthetic pipeline check ready for review: CI now covers both baseline
  architectures, adds a new growth/melt synthetic dynamics, and closes a lint gap and
  a testing gap.
  
  Closes #177. Closes #367.

  ## Changes

  - **CI runs both architectures.** `synthetic_pipeline_check.yaml` is now a matrix over
    the UNet and CNN-ViT-CNN baselines. Each arm runs the check end-to-end (30% improvement
    gate) plus a lenient per-model validation-RMSE threshold. Both fit comfortably in the
    CPU-runner budget (the CNN arm is ~2.5 min: full-capacity model at 48×48, 6 epochs —
    it converges by ~epoch 2).

  - **New `--dynamics grow-shrink` synthetic mode.** Instead of translating a circle, a
    stationary blob grows and shrinks in place via morphological dilation/erosion (an
    open/close cycle), mimicking seasonal ice advance/retreat — exercising concentration
    change in place rather than advection. Default (`moving`) is unchanged. Implemented in
    pure NumPy (no new dependency). Verified it trains end-to-end and passes the learning
    gate on the UNet.

  - **Unit tests for the synthetic module** (`tests/synthetic/`, 52 tests) covering both
    shape generators, the morphology helpers, multi-trajectory stitching, the zarr writer,
    and the pipeline-check helpers (grid validation, split assignment, learning gate).

  - **Lint fixes** in `centroid_error.py` and `base_model.py` (named constant + docstring)
    that were failing `ruff`.

  - Docs updated for the new `--dynamics` option and the two-model CI.

  ## Validation

  `ruff` + `ruff format` clean · `mypy` clean (180 files) · `pytest` 967 passed, 1 skipped ·
  all pre-commit hooks pass.


AoifeHughes:
Should close #367

docs/src/how-to/synthetic-check.md
```diff
@@ -0,0 +1,71 @@
...
+uv run imp synthetic check --config-name baseline/synthetic_unet --dynamics grow-shrink
+```
+
+`moving` tests whether the model learns to translate a fixed shape; `grow-shrink` tests whether it learns concentration change in place (the shape never moves, but its extent pulses). Both work with either baseline and at any valid grid size.
+
+## Notes and constraints
+
+- **Grid sizes must respect each architecture's minimum.**
+  Every grid size must be a multiple of 16 and larger than 16: the UNet processor pools the grid four times, and the check applies this constraint to all models.
+  The ViT processor additionally patchifies the grid with `patch_size: 24`, so for `synthetic_cnn_vit` the grid must also be a multiple of 24 — in practice, use a multiple of 48 (the least common multiple). The default `--grid-size 32` therefore works for the UNet but **not** for the CNN-ViT, whose smallest working grid is 48.
```
jemrobinson: If this is constraining what you want to do, you can change the patch size in the config/on the command line.

icenet_mp/callbacks/loss_history_callback.py
```diff
@@ -0,0 +1,57 @@
...
+import logging
+from pathlib import Path
+
+from lightning import LightningModule, Trainer
+from lightning.pytorch import Callback
+
+logger = logging.getLogger(__name__)
+
+
+class LossHistoryCallback(Callback):
```
jemrobinson: I think this should be a `Logger` not a `Callback`. Would [`CsvLogger`](https://lightning.ai/docs/pytorch/stable/extensions/generated/lightning.pytorch.loggers.CSVLogger.html#lightning.pytorch.loggers.CSVLogger) do what you need?

icenet_mp/cli/main.py
```diff
@@ -30,6 +31,7 @@
 )
 app.add_typer(datasets_cli, name="datasets")
 app.add_typer(evaluation_cli)
+app.add_typer(synthetic_cli, name="synthetic")
```
jemrobinson: If there's only a single command here, it might be better to use a single verb rather than `synthetic check` here (as is done for `train` and `evaluate`). I'll leave it up to you to find one that makes sense!

docs/src/how-to/synthetic-check.md
AoifeHughes: Yeah I was struggling a little with this one because I wanted to be able to run against existing configs as is. But this was a hard constraint so didn't know how best to work around while respecting original values

icenet_mp/synthetic/zarr_writer.py
```diff
@@ -0,0 +1,84 @@
+"""Write synthetic frame sequences as anemoi-compatible zarr stores.
+
+Mirrors the minimal zarr layout that ``anemoi.datasets.open_dataset`` expects (as used
+by ``tests/conftest.py``'s ``build_zarr``), so a synthetic sequence can be consumed
```
jemrobinson: Let's not duplicate this functionality. If it needs to be here, then we don't need `build_zarr` in the test folder.

icenet_mp/models/base_model.py
```diff
@@ -42,6 +43,7 @@ def __init__(  # noqa: PLR0913
         output_space: DictConfig,
         scheduler: DictConfig,
         loss: DictConfig,
+        use_centroid_metric: bool = False,
```
jemrobinson: I think it would be better to handle this by adding a `metrics` config key that defaults to `["accuracy", "mae", "rmse", "sieerror"]` and can be extended/overridden in the config file.

icenet_mp/synthetic/pipeline_check.py
```diff
@@ -0,0 +1,333 @@
...
+    logger.info(
+        "Wrote %d independent synthetic trajectories (%d days total) to %s.",
+        len(dataset.spans),
+        len(dataset.dates),
+        zarr_path,
+    )
+    return zarr_path, dataset.frames, dataset.dates, _split_ranges(dataset.spans)
+
+
+def _add_wandb_logger(config: DictConfig) -> str:
```
jemrobinson: I don't understand this. If you run with the WandbLogger configured normally, it should pick up the WANDB_ENTITY from your local environment. It doesn't seem like a good idea to hard-code it here (especially since we know that people outside Turing are interested in using this codebase).

Also, when do we want this to log to W&B?

.github/workflows/synthetic_pipeline_check.yaml
```diff
@@ -0,0 +1,66 @@
...
+                  f"FAILED: best validation RMSE {best:.4f} exceeds {THRESHOLD}"
+              )
+          EOF
+
+      - name: Upload synthetic check report and plots
+        if: always()
+        uses: actions/upload-artifact@v7
+        with:
+          name: synthetic-pipeline-check
+          path: outputs/synthetic_check/report
```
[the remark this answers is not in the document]
jemrobinson: Do this please.

icenet_mp/metrics/centroid_error.py
```diff
@@ -0,0 +1,82 @@
...
+import torch
+from torchmetrics import Metric
+
+# Frames whose target has less total mass than this are treated as empty (undefined
+# centroid) and excluded from the average; it also floors the denominator so the
+# centroid of an empty frame never divides by zero.
+_EMPTY_MASS_THRESHOLD = 1e-8
+
+
+class CentroidErrorPerForecastDay(Metric):
```
jemrobinson: Can this inherit from BaseErrorMetricDaily?

icenet_mp/models/processors/vit.py
```diff
@@ -68,13 +74,35 @@ def __init__(
 
         self.norm = nn.LayerNorm(emb_dim)
 
-        # (B, N, patch_size * patch_size * out_channels *)
-        self.decoder = nn.Sequential(
-            nn.Linear(emb_dim, patch_size * patch_size * self.out_channels),
+        # Project each patch token to a *feature* map at pixel resolution, not directly
```
jemrobinson: This change should be pulled out into a separate PR. We should compare performance with and without these extra conv blocks.

jemrobinson (changes requested):
Overall, this looks great - I'm interested in trying it out.

However, there's a lot going on here and it might be better to split this into multiple PRs if you can see obvious ways to do so (for example splitting the "local run with synthetic data" from the "CI run with synthetic data"). Definitely the ViT model changes should be split into their own mini-PR so we can separately evaluate the impact.

icenet_mp/models/processors/vit.py
AoifeHughes: ah, I let this slip through. Sorry I was testing a lot of things and thought I'd managed to undo a lot of the model specific tweaks I was testing when I made the PR. sorry!

docs/src/how-to/synthetic-check.md
jemrobinson: I don't think it matters whether the synthetic code tests the exact models we would use for SIC prediction, it's more like a smoke test (in CI at least) so it's fine to alter model parameters to fit the needs of the synthetic dataset.

jemrobinson:
Thinking about this a bit more, I suggest that we don't log this to W&B at all.

1. It doesn't cleanly fit into what we usually mean by train or evaluate so it would need its own project
2. The main purposes of this are for either CI (no output needed) or local running (local plots/videos preferred to W&B).
3. This could simplify some of the logic in the pipeline check

Thoughts @AoifeHughes ?

icenet_mp/synthetic/pipeline_check.py
```diff
@@ -0,0 +1,333 @@
...
+            f"named '{SYNTHETIC_DATASET_NAME}' -- use the 'data=synthetic' config group."
+        )
+        raise ValueError(msg)
+
+    _validate_grid_size(grid_size)
+    trajectories = _make_trajectories(
+        dynamics=dynamics, grid_size=grid_size, n_trajectories=n_trajectories
+    )
+    dataset = generate_multi_trajectory_dataset(trajectories)
+    zarr_path = output_dir / "data" / "anemoi" / f"{SYNTHETIC_DATASET_NAME}.zarr"
```
jemrobinson: If we want this to write to the "anemoi" directory, I think it would be better to structure this as an anemoi data source (see data_processors/sources) steered by a config living in config/data/datasets. This would then mean that we could also test the `datasets create` machinery when we test `synthetic check`.

Does that make sense or would you like to talk this through?

AoifeHughes:
@jemrobinson I think I've addressed your comments in the last few commits.

AoifeHughes:
I think the only thing open is if you wanted to remove the ci and split that to another pr or not. Im not sure it's needed, but happy to do that if you feel its important.

icenet_mp/data_processors/sources/synthetic.py
```diff
@@ -0,0 +1,87 @@
...
+from anemoi.datasets.create.sources import source_registry
+from anemoi.datasets.create.sources.xarray import load_one
+from anemoi.datasets.dates.groups import GroupOfDates
+from earthkit.data import FieldList
+
+from icenet_mp.synthetic.trajectories import generate_default_dataset
+
+
+@source_registry.register("synthetic")
+class SyntheticSource(Source):
```
jemrobinson, line 18: Yes, this is exactly what I meant, thanks!

icenet_mp/metrics/centroid_error.py
```diff
@@ -0,0 +1,55 @@
...
+
+from .daily_metrics import BaseErrorMetricDaily
+
+# Frames whose target has less total mass than this are treated as empty (undefined
+# centroid) and excluded from the average; it also floors the denominator so the
+# centroid of an empty frame never divides by zero.
+_EMPTY_MASS_THRESHOLD = 1e-8
+
+
+class CentroidErrorPerForecastDay(BaseErrorMetricDaily):
```
jemrobinson, line 13: Could this also be useful for non-synthetic data? Doesn't matter if not, but have you thought about it?

icenet_mp/models/base_model.py
```diff
@@ -84,11 +91,25 @@ def __init__(  # noqa: PLR0913
...
+        _metric_classes: dict[str, type[Metric]] = {
+            "accuracy": IceNetAccuracy,
+            "mae": MAEPerForecastDay,
+            "rmse": RMSEPerForecastDay,
+            "sieerror": SeaIceExtentErrorPerForecastDay,
+            "centroid_error": CentroidErrorPerForecastDay,
+        }
+        metric_names = (
+            metrics
+            if metrics is not None
```
jemrobinson, line 103: Do we want to allow users to pass an empty list of metrics? If so, this code is fine.

.github/workflows/synthetic_pipeline_check.yaml
```diff
@@ -0,0 +1,93 @@
...
+      - uses: actions/setup-python@v6
+        with:
+          python-version: "3.11"
+
+      - uses: astral-sh/setup-uv@v7
+        with:
+          version: "0.8.3"
+
+      # Test 1: the pipeline runs end-to-end and passes its improvement gate.
+      - name: Run small synthetic ${{ matrix.name }} check
```
jemrobinson: Should we have a `uv run imp datasets create --config-name ${{ matrix.config }} ${{ matrix.overrides }}` here?

icenet_mp/cli/main.py
```diff
@@ -33,6 +42,107 @@
 app.add_typer(training_cli)
 
 
+@app.command("check")
```
jemrobinson: Can we move this command into a `check.py` in the same folder (as is done for `evaluate`?

icenet_mp/synthetic/pipeline_check.py
```diff
@@ -0,0 +1,327 @@
...
+    _clear_loss_history(metrics_path)
+
+    # Work on a detached copy so we never mutate the caller's config in place.
+    config = DictConfig(OmegaConf.to_container(config, resolve=False))
+    config["base_path"] = str(output_dir)
+    if max_epochs is not None:
+        config["train"]["trainer"]["max_epochs"] = max_epochs
+    config["model"]["encoders"]["latent_space"] = [grid_size, grid_size]
+
+    zarr_path, frames, dates, split_ranges = _generate_dataset(
```
jemrobinson: Instead of this, we should run `uv run imp datasets create --config-name <...>` before calling `uv run imp check ...`. This will create the dataset locally iff it doesn't exist and also means that when we run and end-to-end test in CI we're also testing the `datasets create` path.

This then means that we don't need `_generate_dataset`, `generate_default_dataset` and perhaps some other functions.

jemrobinson (changes requested):
Looks good. Some additional thoughts about how to integrate this into the existing architecture a bit more cleanly.

icenet_mp/metrics/centroid_error.py
AoifeHughes: Yes, and short answer is I think maybe useful. - but it's really a workaround to what I think is a pre-existing problem of the setup with data being unlabelled in the inputs leading to a few issues, which is outside the scope of this and I think I've mentioned before.

jemrobinson: Can you open an issue about the unlabelled data point?

AoifeHughes: Yes I can do, I wanted to have a more concrete understanding ~ it's what some of my long winded rambles on slack have been about and I just need to dedicate some time to detangle. That's also why some random fixes had slipped into this PR initially.

AoifeHughes:
Am I missing something, I'm still seeing a changes requested, but all conversations are resolved?

jemrobinson:
> Am I missing something, I'm still seeing a changes requested, but all conversations are resolved?

"Changes requested" doesn't get dismissed until you request a re-review. I can take a look tomorrow.

icenet_mp/data_processors/sources/__init__.py
```diff
@@ -44,11 +47,17 @@ def register_sources() -> None:
         Union[*_schemas()],
         Discriminator(_discriminator),
     ]
+    _recipe.Action = _action.Action
+    Recipe.model_fields["input"].annotation = _action.Action | None
+    Recipe.model_fields["data_sources"].annotation = (
+        dict[str, _action.Action] | list[_action.Action] | None
+    )
```
jemrobinson: What does this do?

icenet_mp/loggers/local_file_logger.py
```diff
@@ -0,0 +1,119 @@
+"""A Lightning logger that writes images, videos, and metrics to local files.
```
jemrobinson, line 1: Did you replace this with `CsvLogger`? If so, we can remove this file.

If not, can we move this to a top-level `loggers` directory (i.e. at `icenet_mp/loggers`)? This would then be equivalent to our `callbacks` directory.

icenet_mp/model_service.py
```diff
@@ -272,15 +272,20 @@ def build_trainer(  # noqa: C901
...
-                logger_config,
-                job_type="multistage" if job_stage else "single-stage",
-                project=project,
-            )
-            for logger_config in self.config.get("loggers", {}).values()
-        ]
+        # Setup Lightning loggers — only pass job_type/project to W&B loggers.
+        extra_loggers = []
+        for logger_config in self.config.get("loggers", {}).values():
+            is_wandb = "wandb" in str(logger_config.get("_target_", "")).lower()
```
jemrobinson: Maybe this is more robust?

```suggestion
            is_wandb = logger_config.get("_target_", "").split(".")[-1] == "WandbLogger"
```

icenet_mp/loggers/local_file_logger.py
```diff
@@ -0,0 +1,119 @@
...
+
+Implements the subset of the `WandbLogger` interface used by `PlottingCallback`
```
jemrobinson, line 3: When do you want to enable the local logger? Is this only for runs with WANDB_MODE=offline? Because whenever you're running with W&B synchronisation on, you already get a local copy of the files in `wandb/run-<whatever>/files/media/images`.

jemrobinson (changes requested):
A few minor changes - mostly about moving things around. LGTM though, so happy to approve once these are done.

docs/src/user-guide/commands.md
```diff
@@ -61,6 +74,8 @@ Checkpoints are saved to `${BASE_DIR}/training/wandb/run-<date>-<id>/checkpoints
 uv run imp evaluate --checkpoint PATH_TO_A_CHECKPOINT
 ```
 
+To evaluate a synthetic run, add `--config-name synthetic`.
```
jemrobinson: This isn't really specific to synthetic runs. There's a default value for `--config-name` so any time you're running without specifying `--config-name` you use the default.

docs/src/user-guide/commands.md
```diff
@@ -39,6 +45,13 @@ Trains the model end-to-end:
 uv run imp train
 ```
 
+Synthetic experiments do not use W&B. Use the synthetic configuration, which
+saves metrics and plotting artefacts locally under `${BASE_DIR}/report`:
+
+```bash
+uv run imp train --config-name synthetic
```
jemrobinson, line 52: This crashes with:

```
> uv run imp train --config-name synthetic
😈 [2026-07-29 14:21:01] Anti-aliasing disabled for compatibility with deterministic running and/or accelerator architecture.
😈 [2026-07-29 14:21:01] Building a new 'icenet_mp.models.EncodeProcessDecode' model...
😈 [2026-07-29 14:21:01] Found 1 dataset groups.
😈 [2026-07-29 14:21:01] 1) sic-icenet:
😈 [2026-07-29 14:21:01]    - /Users/jrobinson/Developer/forecasting/icenet/base/data/anemoi/synthetic-sic.zarr
...
FileNotFoundError: active mask is requested but no mask was found at ../base/data/preprocessing/masks/synthetic-sic/active_mask.npy. Masks are generated for SSMIS datasets during `datasets create`.

The above exception was the direct cause of the following exception:
...
InstantiationException: Error in call to target 'icenet_mp.models.encode_process_decode.EncodeProcessDecode':
InstantiationException("Error in call to target 'icenet_mp.models.decoders.naive_linear_decoder.NaiveLinearDecoder':\nFileNotFoundError('active mask is requested but no mask was found at ../base/data/preprocessing/masks/synthetic-sic/active_mask.npy. Masks are generated
for SSMIS datasets during `datasets create`.')")
full_key: model
```

You might need to generate some dummy masks for compatibility in `data_downloader.py`.

jemrobinson:
Ran a `datasets create` and a `train` job and got these results. Are they as expected?

00005__output_static__2020-05-27-sea-ice-concentration_0
00011__output_static__2020-05-27-sea-ice-concentration_0

AoifeHughes:
> Ran a `datasets create` and a `train` job and got these results. Are they as expected?
> 00005__output_static__2020-05-27-sea-ice-concentration_0 00011__output_static__2020-05-27-sea-ice-concentration_0

Yes. It's also why I spent a lot of time before making the PR suitable fixing and trying to solve the problems and lead me to thinking it's the actual models themselves are flawed... but that's a whole other issue.  I think the issue you're seeing here was one of the ones that was fixed by labelling the inputs with an order rather than not being explicit.

docs/src/user-guide/commands.md
jemrobinson: Done in 6a50e9d

docs/src/user-guide/commands.md
jemrobinson: Done in e3ecec4

icenet_mp/loggers/local_file_logger.py
jemrobinson: Done in dee605a

icenet_mp/data_processors/sources/__init__.py
AoifeHughes: It's a clunky way to get around data source rejections for the synthetic data type. Happy for alternative solutions. But was trying to not disrupt core things

icenet_mp/model_service.py
AoifeHughes: this was applied in one of the commits yesterday, thanks.

icenet_mp/loggers/local_file_logger.py
AoifeHughes: After this and some other comments I thought it best to make the assumption you'd never want wandb for these models.

jemrobinson (approved):
LGTM

AoifeHughes:
Thanks James!

PR #377, opened 2026-07-14
Fix multistage slowdown

### LINKED ISSUE #376: Fix slowdown in the final stage of --multistage training
### Problem / Goal

`imp train --multistage` trains an `EncodeProcessDecode` model in one continuous process
across sequential stages (4 encoders, decoder, processor, finetune). The `finetune` stage is very slow when run in this way. However, running the same `finetune` stage in a standalone process (after loading previous stage checkpoints) was 30x faster.

This likely due to a memory leak somewhere.

### What does good look like?

- [ ] Root cause confirmed: identify specifically what accumulates across stages (leaked DataLoader worker processes, MPS allocator fragmentation, retained `Trainer`/model references, or something else)
- [ ] Fix for `train_multistage`, e.g. releasing each stage's resources before starting the next stage and/or clearing caches between stages

### Related context

Workaround in the meantime: run each multistage stage as a **separate** `imp train --multistage`  invocation, chaining with `--checkpoint-dir` to skip completed stages.

jemrobinson:
Clear device caches to mitigate multistage slowdown.

Closes #376

### DIFF
--- a/icenet_mp/model_service.py
+++ b/icenet_mp/model_service.py
@@ -1,3 +1,4 @@
+import gc
 import logging
 import os
 from pathlib import Path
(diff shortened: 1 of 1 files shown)


Shaerdan (approved):
LGTM

PR #265, opened 2026-05-07
Model card

aranas:
Adding a first draft for a TRL 2 mode card following IICWG templates: https://github.com/IICWG/AI-ML-TRLs

This card provides a description of the current research stage of the project as well as advanced (TRL3) stage of the codebase.

One major information that is missing but required for the model card is a versioned release number associated.

PR #380, opened 2026-07-15
:memo: Update model card to match IICWG TRL TT and add in release version

louisavz:
Minor updates to model card to match the one pushed to IICWG TRL task force test repo and add in release version as tagged last week.

This addresses the missing versioned release number mentioned in #265


jemrobinson (approved):
LGTM

PR #390, opened 2026-07-23
Add AMSE spectral loss for anti-blur (#304)

### LINKED ISSUE #304: Implement a loss function that avoids the double penalty
### Problem / Goal

Loss functions for forecasting need to avoid double penalty in prediction to avoid over-smoothing. See e.g. https://arxiv.org/abs/2501.19374

### What does good look like?

- [ ] Better loss function implemented (e.g. Spherical Loss)

### Related context

- Talk to Cumulus and FastNet teams about what they do

Shaerdan:
Closes #304  This is a flat domain variance of the Adjusted spherical-harmonic MSE (AMSE) of Subich et al. (2025), arXiv:2501.19374. 

This variance uses the FFT-spectral annular-bin power/coherence terms with the original paper's maximum decorrelation weighting, and in addition a hybrid mode (Huber + weighted AMSE excess) that reduces to the default Huber loss at spectral_weight=0. Selectable via 'imp train loss=amse' (or config file loss=amse).

Implementation notes:
- Scale groups: annular bins of isotropic wavenumber |k| from a 2D DFT, note: on the fixed area EASE2 grid, Parseval's identity makes the spectral decomposition of MSE exact, the boundary ring is inactive (both 0 in prediction and in target), so the DFT periodicity assumption wont introduce boundary discontinuity.
- Two modes: (a) hybrid (default) = Huber + spectral_weight * (AMSE − MSE), returns to exactly to the default Huber loss at `spectral_weight=0`;  (b) pure = Eq (6) of  Subich et al. (2025).
- Numerical guards and other computational details mirror the reference implementation (the original github repo of AMSE) with some adaptations.
- One class, one config file; no changes to the training code (uses the loss config group from #283). Scope: spatial-field predictions; untested with losses computed in other spaces (e.g. DDPM v-space).

## Testing
- New tests/models/test_amse_loss.py (13 tests): Hydra instantiation; fail-loud arg validation; loss(x,x)=0; Parseval reconstruction of the spatial MSE on even and odd grids (pins the FFT/binning/factor-2 bookkeeping); blur penalized monotonically; amplitude gradient points upward for damped predictions; finite gradients for all-zero fields; masked-field sanity; symmetry; exact Huber spectral_weight=0; bf16 input safety.
- tests/models/test_loss_config.py: amse added to the param
- Full suite: 941 passed, 1 skipped. pre-commit run --all-files and mypy .clean.
- End-to-end: composes and trains with --config-name=baseline/ predict=sic-ssmis-14d; a 50-epoch A/B against `loss=huber` (same seed, early stopping disabled in both arms) is running spectral PSD ratio and coherence against lead day) to follow on the issue.

### DIFF
--- a/icenet_mp/losses/amse_loss.py
+++ b/icenet_mp/losses/amse_loss.py
@@ -1,0 +1,224 @@
+"""Adjusted-MSE (AMSE) spectral anti-blur loss on a flat grid.
+
+Adaptation of the spherical-harmonic AMSE loss of Subich et al. (2025),
+"Fixing the Double Penalty in Data-Driven Weather Forecasting Through a
+Modified Spherical Harmonic Loss Function" (arXiv:2501.19374), to the flat
+equal-area EASE2 grids used by icenet-mp. Independent reimplementation of the
+published formula; reference implementation (spherical, JAX) at
+https://github.com/csubich/graphcast (branch ``amse``).
+
+Pointwise squared-error losses reward damping the amplitude of scales the
+model cannot predict exactly (the "double penalty"): per scale group, the
+MSE-optimal amplitude ratio equals the coherence with the target, so partially
+predictable fine scales are systematically blurred. AMSE removes this
+incentive. Writing per scale group k the powers P_k(x), P_k(y) and coherence
+Coh_k(x, y) of prediction x and target y,
+
+    MSE(x, y)  = sum_k [ (sqrt(P_k(x)) - sqrt(P_k(y)))^2
+                         + 2 * sqrt(P_k(x) * P_k(y)) * (1 - Coh_k(x, y)) ]
+    AMSE(x, y) = sum_k [ (sqrt(P_k(x)) - sqrt(P_k(y)))^2
+                         + 2 * max(P_k(x), P_k(y)) * (1 - Coh_k(x, y)) ]
+
+so under AMSE the decorrelation penalty can no longer be reduced by shrinking
+the predicted amplitude, and the amplitude optimum at any fixed coherence is
+P_k(x) = P_k(y): the target's power spectrum is preserved scale by scale.
+
+On the flat grid the orthonormal decomposition is the 2D DFT and the scale
+groups are annular bins of the isotropic wavenumber |k| = sqrt(kx^2 + ky^2);
+Parseval's identity makes the group decomposition of MSE exact, which is the
+only property the construction needs (Subich et al., Sec. 2.2 and 4.2).
+
+Scope: designed and tested for spatial-field predictions (single-stage
+encode-process-decode). Using this loss with models whose training loss lives
+in a different space (e.g. DDPM v-space) is untested.
+"""
+
+import torch
+from torch import nn
+from torch.nn import functional
+
+# Type alias for the cached per-(H, W, device) binning tensors:
+# (flat mode->bin index, flat mode weights, indices of non-empty bins)
+BinCache = tuple[torch.Tensor, torch.Tensor, torch.Tensor]
+
+# Fields must carry at least (batch-like, H, W) dimensions.
+MIN_FIELD_NDIM = 3
+
+
+class AMSELoss(nn.Module):
+    """Adjusted-MSE (anti-double-penalty) spectral loss on a flat grid.
+
+    Two modes:
+
+    - ``"hybrid"`` (default): the repo's standard Huber loss plus
+      ``spectral_weight`` times the AMSE *excess*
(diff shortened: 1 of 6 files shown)


icenet_mp/losses/amse_loss.py
```diff
@@ -0,0 +1,224 @@
...
+    geometric-mean square root (the coherence denominator is singular for a
+    zero field), coherence clamped at 1 from above (Cauchy-Schwarz can be
+    violated by the eps), the mean/DC component handled as a direct squared
+    difference (catastrophic cancellation), and all spectral arithmetic in
+    float32.
+    """
+
+    def __init__(
+        self,
+        mode: str = "hybrid",
```
jemrobinson:
```suggestion
        mode: Literal["hybrid", "pure"] = "hybrid",
```

Not essential, but if you do this, you also need `from typing import Literal`

jemrobinson (changes requested):
This looks great - I haven't double-checked the maths here but it looks sensible to me.

Could you show some results comparing a training run using `HuberLoss` against one using `hybrid` and another using `pure`? It would be great if you should actually show a reduction in blurring on an evaluation run too.

icenet_mp/losses/amse_loss.py
```diff
@@ -0,0 +1,224 @@
...
+        geo_mean = torch.sqrt(self.eps + power_p * power_t)
+        coherence = torch.clamp(cross / geo_mean, max=1.0)
+        excess_weight = (torch.maximum(power_p, power_t) - geo_mean).clamp_min(0.0)
+        return (2.0 * excess_weight * (1.0 - coherence)).sum(dim=1)
+
+    def pure_amse(self, prediction: torch.Tensor, target: torch.Tensor) -> torch.Tensor:
+        """Per-field AMSE (amplitude + adjusted decorrelation terms + DC term)."""
+        power_p, power_t, cross, dc = self.binned_spectra(prediction, target)
+        geo_mean = torch.sqrt(self.eps + power_p * power_t)
+        coherence = torch.clamp(cross / geo_mean, max=1.0)
```
jemrobinson, line 360: Could potentially factorise this into a function used by both `pure_amse` and `spectral_excess` but since only one of these will be called in a given run, it doesn't seem that important. The only benefit is to ensure that they stay in sync.

jemrobinson:
There's a lot of information about pros/cons of the various loss functions that currently lives as comments in the YAML files. This PR could be a good place to move that into a short docs page explaining:

- how to change which loss function to use for training
- which loss functions we support and the pros/cons of each

This doesn't have to be new content, just moving things that we already have.

Shaerdan:
> This looks great - I haven't double-checked the maths here but it looks sensible to me.
> 
> Could you show some results comparing a training run using `HuberLoss` against one using `hybrid` and another using `pure`? It would be great if you should actually show a reduction in blurring on an evaluation run too.

Good idea, will be done before Monday 3rd of Aug, currently setting things up on JASMIN and will start testing on the real data. I have failed to set up a nice toy example so far and will continue be working on it, hopefully can have both real data results and a suitable toy results by Monday

Shaerdan:
> This looks great - I haven't double-checked the maths here but it looks sensible to me.
> 
> Could you show some results comparing a training run using `HuberLoss` against one using `hybrid` and another using `pure`? It would be great if you should actually show a reduction in blurring on an evaluation run too.

I attached some results here comparing huber loss against hybrid and pure. 

here is the setup I used for this comparison: 
{default rollout, residual rollout (#409)} × {Huber, AMSE hybrid}  all combination tested, each trained with 40 epochs, same seed, identical recipe; spectral_weight tuned per architecture to a be 0.1 at initialisation. In terms of sharpness, hybrid wins in this toy example test, but i notice MAE increased a little with the hybrid and AMSE compared to the blurry Huber result. Could mean more fine tuning is required or could be that the extra details retrained necessarily bring in more noise. Need further investigations

sharpness-by-lead-incl-pure

one important note i'd add: since the production models available back then turned out to have an architectural issue (low capacity to capture temporal changes), it masked any potential improvements brought by the new AMSE loss, and gave raise to a separate issue (#409). another note: the controlled comparison now run on a toy example/synthetic data, designed to capture the temporal change on a scale similar to the actual data, with sharp features used as markers to visually check the anti-blur effects. 

It's a good news to share that the new residual learning architecture with feedback markovian rollout + new AMSE loss combination beats persistence by some distances, at least in this toy example test. Ive been trying this right now on the actual SIC full dataset, and will add results once they are in, but i think i will add them to issue #409 instead (that's the one specifically dedicated to the new residual learning architecture).

jemrobinson (approved):
LGTM after seeing results on #409

Issue #169, opened 2025-10-09
Add new contour plot that visualizes sea ice edge


aranas:
### Problem / Goal

We currently do not indicate the sea ice edge (SIC 15% treshhold) in our plots.
When we evaluate model performance through integrated ice edge error we may want to provide plots that help us compare predicted and observed ice edge as a visual accompanying the IIEE metric.

### What does good look like?

- [ ] user has plot_ice_edge option in configuration
- [ ] if User chooses to plot the ice edge they get an additional plot that visualizes a map of the SIC metric with a contour drawn at SIC 15% (or user configurable treshold) as well as a benchmark contour based on the SIC 15% boundary from the observed data

### Related context

- see sea ice edge viz examples in this issue alan-turing-institute/ice-station-zebra-project-board#51

jemrobinson:
Done in #396

PR #396, opened 2026-07-27
Adding additional evaluation metrics

### LINKED ISSUE #125: Add additional metrics to evaluation
### Problem / Goal

We want to evaluate the model on at least the following metrics:
- accuracy (for example binary accuracy as in Andersson 2021), see #113 
- integrated ice edge error
- SSIM

Also consider:
- latitude-weighted Root Mean Square Error
- Metrics recommended by Melsom et al (2019) https://os.copernicus.org/articles/15/615/2019/
- Metrics uesd by Palerme et al (2025) https://egusphere.copernicus.org/preprints/2025/egusphere-2025-2001/egusphere-2025-2001.pdf
- Adding in a visualisation of the sea ice edge contour #169 
- Fractions skill score (FSS)
- Ensemble metrics:
   - Rank histograms
   - Crps
   - Spread skill ratio
   - Anomaly coefficients (e.g. https://www.nature.com/articles/s41612-025-01058-0)

### What does good look like?

- [x] At test, predicted SIC will be translated into different metrics
- [x] accuracy curves plotted & saved in W&B #113
- [x] integrated ice edge error computed and saved to W&B

### Related context
- #113 
- #169
- #187

IFenton:
Closes #125, #169, #469


IFenton:
Closes https://github.com/alan-turing-institute/icenet-mp/issues/169

docs/src/user-guide/metrics.md
```diff
@@ -0,0 +1,246 @@
...
+`SSIMPerForecastDay` compares fields within a Gaussian-weighted window, combining
+local luminance, contrast, and structure (Wang et al., 2004), rather than comparing
+every pixel independently.
+
+**Limitations:** SSIM is the one metric here that penalises `noisy` heavily (SSIM ≈
+0.58 averaged over the forecast) despite its low MAE/RMSE and high accuracy — independent per-pixel noise
+disrupts the local variance/covariance structure SSIM measures even though it
+averages out to a small mean error. It broadly agrees with most metrics that `persistence`
+and `shifted edge` are worst overall (both systematically wrong at every scale), but
+the disagrees sharply on `noisy`, which most metrics rate as one of the *best* scenarios.
```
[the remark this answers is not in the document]
IFenton: Resolved by [7b26bc5](https://github.com/alan-turing-institute/icenet-mp/pull/396/commits/7b26bc50f77ade6cd3ed01a9a824faf2d6ffd2ba)

icenet_mp/metrics/fss.py
```diff
@@ -0,0 +1,178 @@
...
+            self.count += batch_count
+
+    def compute(self) -> torch.Tensor:
+        """Compute the final FSS per lead time."""
+        if self.count.numel() == 0:
+            return torch.tensor([], dtype=torch.float32, device=self.sum_mse.device)
+
+        mean_mse = self.sum_mse / self.count
+        mean_mse_ref = self.sum_mse_ref / self.count
+        return 1 - mean_mse / mean_mse_ref
```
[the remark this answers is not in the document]
IFenton: Resolved in [95a72e3](https://github.com/alan-turing-institute/icenet-mp/pull/396/commits/95a72e3ae5aafaad2f3431c6d613a571ab529a2f)

icenet_mp/models/base_model.py
```diff
@@ -96,13 +119,39 @@ def __init__(  # noqa: PLR0913
         self.lr_scheduler_cfg = lr_scheduler
         self.loss_cfg = loss
 
+        # Land mask for ice-edge metrics (excludes land/ice boundaries from FSS/DIIEE)
+        land_mask: torch.Tensor | None = None
+        if mask_dir is not None:
```
AoifeHughes, line 124: Is there a test for this, if Im reading right it could mean that this is passed for every model_service call, and if it is not none then a file not found is raised. So any dataset without a landmask would crash?

icenet_mp/metrics/fss.py
```diff
@@ -0,0 +1,185 @@
...
+
+        Parameters
+        ----------
+        lambda_truth : torch.Tensor
+            Local edge-cell fraction field for the target, shape (N, H, W).
+        lambda_model : torch.Tensor
+            Local edge-cell fraction field for the prediction, shape (N, H, W).
+
+        """
+        n = self.neighborhood_size
```
AoifeHughes, line 107: This averages MSE across nxn offset sub-grids with an unweighted mean, but if h/q aren't divisable by neighborhood size (neighbour? irks me that code uses non-British English but thats not part of this, I'm just waffling). So does this biases FSS from the true grid MSE? I'm not sure, but something to think about

icenet_mp/models/base_model.py
```diff
@@ -112,6 +161,11 @@ def __init__(  # noqa: PLR0913
                 "mae",
                 "rmse",
                 "sieerror",
+                "iiee",
+                "diiee",
+                "centroid_error",
```
AoifeHughes, line 166: given this is part of the default, if Im reading right this would need explicitly to be turned off. I like it as a metric tbh, but do we want to update doc strings / mention somewhere? I mean this is small enough it probably doesn't matter.

icenet_mp/metrics/ssim.py
```diff
@@ -0,0 +1,118 @@
...
+
+
+def _gaussian_kernel(filter_size: int, filter_sigma: float) -> torch.Tensor:
+    """1D Gaussian filter of the given size and standard deviation, normalized to sum to 1."""
+    coords = torch.arange(filter_size, dtype=torch.float32) - filter_size // 2
+    filt = torch.exp(-0.5 * (coords / filter_sigma) ** 2)
+    return filt / filt.sum()
+
+
+class SSIMPerForecastDay(BaseErrorMetricDaily):
```
AoifeHughes, line 23: Test coverage?

AoifeHughes (dismissed):
Looks good, I think the mask_dir thing is the only thing I think needs actually addressed. The FSS stuff is a maybe because maybe I've missunderstood the grid logic, anything else is just thoughts and can be ignored without causing offense to me :)

sylvesterkaczmarek:
There is a multi-target correctness issue in the sea-ice-specific metrics. `BaseModel` applies SIE error, IIEE, DIIEE and FSS to the complete prediction tensor, while those implementations threshold every channel at `SEA_ICE_THRESHOLD` and then sum/reshape across the channel dimension.

That works for a single SIC target, but IceNet-MP can expose multiple target variables. In a configuration such as `[ice_conc, ice_thickness]`, the ice-thickness channel would also be thresholded at `0.15` and contribute to the reported ice extent/edge metrics, even though those metrics are defined specifically for SIC.

Could these metrics either operate explicitly on the SIC target channel, or fail validation when more than the supported SIC channel is passed? A two-channel regression would make the intended contract clear.

sylvesterkaczmarek (commented):
the sea-ice-specific metrics still need an explicit SIC-channel contract. Applying SIE, IIEE, DIIEE and FSS to the complete target tensor means any additional target variable is also thresholded at 0.15 and contributes to the score. I would either select the SIC channel explicitly or reject unsupported multi-target input, ideally with a two-channel regression test.

sylvesterkaczmarek (commented):
One fresh current-head CI blocker: all Python versions fail the two `train_stage_processor` unit tests because this PR now reads `self.data_module.mask_directory` in both processor-stage construction paths, while those focused tests construct `ModelService` with `__new__` and do not initialise `data_module_`. The failing line is the new `mask_dir=str(self.data_module.mask_directory)` addition. The production path has `data_module_`, so this looks like a fixture update rather than a reason to remove the new argument: initialise a mocked `service.data_module_` with `mask_directory` in both processor-stage tests (or use the existing service fixture) and rerun. Separately, my earlier SIC-only/multi-target metrics concern is still unresolved on this head.

icenet_mp/models/base_model.py
jemrobinson: Fixed in 41f28f8e

icenet_mp/metrics/fss.py
jemrobinson: Fixed in 60b44d1e by keeping track of the area actually covered and dividing by it at the end.

icenet_mp/models/base_model.py
jemrobinson: Moved to config file in 2eeb7074

icenet_mp/metrics/ssim.py
jemrobinson: Done in b00f19bc

jemrobinson:
Multi-target issues addressed in 2f63251e

jemrobinson (approved):
LGTM

icenet_mp/metrics/fss.py
```diff
@@ -0,0 +1,195 @@
...
+                ).pow(2).sum(dim=(-2, -1))
+                mse_ref_term = torch.minimum(term1, term2)
+                mse_sum = mse_term if mse_sum is None else mse_sum + mse_term
+                mse_ref_sum = (
+                    mse_ref_term if mse_ref_sum is None else mse_ref_sum + mse_ref_term
+                )
+                total_area += area
+        mse = torch.tensor(0) if mse_sum is None else mse_sum / total_area
+        mse_ref = torch.tensor(0) if mse_ref_sum is None else mse_ref_sum / total_area
+        return mse, mse_ref
```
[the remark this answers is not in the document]
jemrobinson: Done in 056050d2

pyproject.toml
```diff
@@ -76,6 +76,7 @@ notebooks = [
     "download-toolbox>=0.0.6",
     "jupyter>=1.1.1",
     "s2sphere>=0.2.5",
+    "gdal==3.13.0", # must match the system GDAL library (`gdal-config --version`)
 ]
```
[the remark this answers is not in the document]
jemrobinson: Done in a0aa8552

zensical.toml
```diff
@@ -13,8 +13,9 @@ nav = [
         { "Installation" = "user-guide/installation.md" },
         { "Configuration" = "user-guide/configuration.md" },
         { "Data" = "user-guide/data.md" },
-        { "Loss functions" = "user-guide/loss-functions.md" },
         { "Commands" = "user-guide/commands.md" },
+        { "Loss functions" = "user-guide/losses.md" },
```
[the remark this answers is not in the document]
jemrobinson: Done in c211781c

docs/src/user-guide/metrics.md
```diff
@@ -0,0 +1,174 @@
...
+```
+
+or in a config file:
+
+```yaml
+reporting:
+  metrics:
+  - accuracy
+  - sieerror
+  - fss_neighbourhood_5
```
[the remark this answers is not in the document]
jemrobinson: Done in 091a0722

docs/src/user-guide/metrics.md
```diff
@@ -0,0 +1,174 @@
...
+```
+
+or in a config file:
+
+```yaml
+reporting:
+  metrics:
+    - accuracy
+    - sieerror
+    - fss_neighbourhood_5
```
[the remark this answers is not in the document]
jemrobinson: Done in 23ff174b

PR #403, opened 2026-07-28
Update installation.md with explicit link to 'uv' installation

IFenton:
Added installation instructions link for 'uv'.

Very small change as I spent a while trying to figure out how to install uv on an HPC

### DIFF
--- a/docs/src/user-guide/installation.md
+++ b/docs/src/user-guide/installation.md
@@ -2,7 +2,7 @@
 
 ## Prerequisites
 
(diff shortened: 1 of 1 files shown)


jemrobinson:
Isn't this what the existing link points to? Worth removing the duplicate?

IFenton:
> Isn't this what the existing link points to? Worth removing the duplicate?

Yes it was, I just didn't notice that it was a (relevant) link. I've now removed the duplicate

jemrobinson (approved):
LGTM

PR #410, opened 2026-08-03
Add opt-in physical-space rollout with residual (tendency) prediction (#409)

### LINKED ISSUE #409: Add opt-in rollout at physical-space with residual (tendency) prediction (#412)
Refer to PR #405 as the original, i believe there are some overlaps, the combining this to #405 requires some re-organising first.

### Problem

Two following two structural properties of the current EncodeProcessDecode forecast loop
potentially make it hard for to beat the persistence baseline  (this becomes an issue when testing AMSE loss #304 ):

1. **lacks identity fallback.** for the model to forecast "no change" it must reproduce its
   input through a downsampling encoder, the processor bottleneck, and an upsampling
   decoder. On a synthetic benchmark a trained `cnn_vit_cnn`-style configuration
   scored active-cell MAE 0.093 at lead 1 where literally copying the input scores
   0.034 — the model is ~2.7× worse than a copy at the one behaviour every forecast
   starts from. Persistence, the baseline to beat, is effectively outside the
   hypothesis space.
2. **unconstrained latent feedback.** the processor's raw latent output is appended
   to its own input window without ever being decoded, re-encoded, or tied to the
   distribution the encoders produce; in multi-input configurations the fed-back
   latent also carries channels for input groups (weather, ocean profiles) whose
   future values are never supervised. The cheapest way for training to control the
   resulting compounding is a map that changes little per step — consistent with
   forecasts whose total change across leads is half or less of reality's.

This is also related to #272 again but is a different mechanism:
it persists with the windowed rollout that resolved #272, the original cycling defect.

### Proposed change (opt-in, defaults untouched)

Three config flags on EncodeProcessDecode:

- `rollout_space: physical` — close the forecast loop in observation space: each
  step encodes the current window of physical frames, takes one processor step,
  decodes, and rolls the updated physical field into the window (re-encoded next
  step). Non-target input groups hold their last observed frame.
- `predict_residual: true` — the decoder output is a signed daily increment:
  `x_{k+1} = mask(clip(x_k + delta_k, 0, 1))`. The identity forecast is
  `delta = 0`, so persistence becomes the zero point of the hypothesis space.
- `zero_init_tendency: true` — zero the final decoder convolution at
  initialisation, so an untrained model reproduces (masked) persistence bit-exactly
  at every lead and training starts on the baseline instead of below it.

With the flags absent the class is byte-for-byte unchanged.

### Tested

On a gated synthetic advection benchmark (built so that the best possible forecast
is sharp and beats persistence 3.9–5.6×), a 2×2 factorial {default rollout, physical
residual rollout} × {Huber, AMSE (#304)} gives, as MAE ratio to persistence at
leads 1/3/5/7 (below 1.0 beats persistence; train / validation / test splits):

| arm | train | validation | test |
|---|---|---|---|
| physical residual + AMSE | 0.95/0.75/0.71/0.66 | 0.99/0.86/0.92/0.95 | 0.99/0.86/0.87/0.88 |
| physical residual + Huber | 1.27/0.72/0.62/0.59 | 1.18/0.78/0.83/0.89 | 1.34/0.82/0.79/0.81 |
| default rollout + AMSE | 1.26/0.87/0.82/0.74 | 1.43/1.03/1.05/1.02 | 1.21/0.89/0.87/0.83 |
| default rollout + Huber | 1.29/0.89/0.83/0.75 | 1.45/1.03/1.04/1.00 | 1.23/0.90/0.88/0.84 |

The physical-residual configurations also keep 96–100 % of the target fine-resolution-band
spectral power under the AMSE loss, versus ~0.30–0.37 for the default rollout under
either loss. The AMSE formulation makes it a bit easier to see where the loss of details come 
from and quantifies it.

### Comparable costs

The original latent rollout is cheap per step and this proposal does not change
that, one full training step on a 432² / 14-lead / 3-group configuration measured
3.3× slower with the physical loop (with these 1-layer encoders; the gap was hidden
by data loading in an end-to-end A100 comparison, but it is real). The following is noted for 
the stvit future works: the physical variant is also strictly Markovian in the physical window, 
it cant carry latent memory between forecast steps. A latent space repair with the same
persistence representability property (maybe a re-encoding consistency term) would
enable this proposal there, this remains open.

### Notes

- One DDP interaction surfaced while testing at scale: the model holds parameters
  that receive no gradient in some configurations (will be uploaded soon), which errors 
  default DDP; strategy=ddp_find_unused_parameters_true is the workaround been used so far.
- Relates to: #272 (rollout), #304 / #390 (AMSE loss issue, note that the loss and this change are
  complementary but separate issues)

Shaerdan:
Closes #409

In toy example v4 (will be added along side this PR), when we tested it for AMSE loss efficacy, we observed an spin-off issue that the latent rollout predicts a temporally slow varying SIC, measured as roughly 30% of the ground truth temporal variability. This motivates an alternative rollout that uses encoder consistently as a physical space data constraint during the training (something like q(z_i | u_i ), where z_i is the latent vector at step i and u_i is the physical space correspondence). In the toy example  the physical rollout version tracks the ground truth temporal variability much better (near 90% matching in temporal variability). We note that despite the clear improvement in the toy v4, when we tested this in the real data, this becomes less clear (needs more updates and further investigation). Below is the description and design: 

**Setting.** Input groups g ∈ {era5, float-argo, sic-ssmis}; target group g\* = sic-ssmis.
Physical grid 432×432, latent grid 216×216. Latent channels, measured: era5 56 + argo 4 +
sic 12 = **c = 72**; the decoder emits **1** channel. History n_h = 3, forecast n_f = 14. We denote a timestep of latent inputs (that has 72 channels) as a latent frame.

### notation table

| symbol | type / shape (`full_south`) | meaning |
|---|---|---|
| G | {era5, float-argo, sic-ssmis} | the input groups |
| g\* | = sic-ssmis | the target group: the one being forecast |
| C_g | era5 28, argo 2, sic-ssmis 6 | physical channels of group g |
| H×W | 432×432 | physical grid |
| h×w | 216×216 | latent grid |
| c_g | era5 56, argo 4, sic-ssmis 12 | latent channels of group g (measured) |
| c | 72 | total latent channels, c = Σ_g c_g |
| n_h, n_f | 3, 14 | history steps in, forecast steps out |
| p | p ∈ {1,…,n_h} | position in a window; **p = n_h is the newest**, i.e. the forecast origin |
| k | k ∈ {1,…,n_f} | forecast step (lead) |
| j | j = 1 (0-based) | index of `ice_conc` within g\*'s C_{g\*}=6 channels (`feedback_channel`) |
| x_g[p] | ℝ^{C_g×H×W} | observed physical frame of group g at window position p |
| E_g | ℝ^{C_g×H×W} → ℝ^{c_g×h×w} | encoder for group g, applied to ONE frame |
| Φ | ℝ^{(n_h·c)×h×w} → ℝ^{c×h×w} | processor: n_h latent frames in, ONE latent frame out |
| D | ℝ^{c×h×w} → ℝ^{1×H×W} | decoder, raw output |
| R, M | ℝ^{1×H×W} → ℝ^{1×H×W} | range restriction (identity when `restrict_range: none`); active-cell mask |
| ŷ_k | ℝ^{1×H×W} | the model's forecast for lead k |

Two constructions used throughout:

    cat_g( · )   concatenate over the groups g ∈ G along the channel axis  → c channels
    cat_p( · )   concatenate over the window positions p = 1..n_h along the channel axis
                                                                          → n_h·c channels

Encoding one window position, for every group at once:

    z[p] = cat_g ( E_era5(x_era5[p]) , E_argo(x_argo[p]) , E_sic(x_sic[p]) ) ∈ ℝ^{c×h×w}

Finalisation of a raw decoder output r against an anchor a (this is `finalise()`),
`base_decoder.py:82-113`, for an additive skip connection:

    F(r, a) = M( clamp( R(r) + a , 0 , 1 ) )

Finalisation with an additive skip and anchor a (`base_decoder.py:82-113`):

    F(r, a) = M( clamp( R(r) + a , 0 , 1 ) )     M = active-cell mask, R = range restriction

**(a) Latent rollout (main's default)** (`base_processor.py:88-99`)

The window holds latent frames. Every entry of W_0 is an encoder output:

    W_0 = ( z[1] , z[2] , … , z[n_h] )

    for k = 1 … n_f:
        ζ_k = Φ( cat_p( W_{k-1} ) )                       ζ_k ∈ ℝ^{c×h×w}
        W_k = ( W_{k-1}[2] , … , W_{k-1}[n_h] , ζ_k )     drop the oldest, append ζ_k

`ζ_k` is appended **without any encoder being applied to it**. Hence for k ≥ n_h every entry
of W_k is a processor output rather than an encoder output.

The anchor is a single frame, chosen once and independent of k:

    a = x_{g*}[n_h] [ j : j+1 ]        one channel: ice_conc of the newest observation

    for k = 1 … n_f:
        ŷ_k = F( D(ζ_k) , a )

**(b) Physical rollout (what this PR adds)** (`encode_process_decode.py:325-351`)

Only the **target** group (sic) is rolled forward, all other groups are fixed at the forecast
origin, its window filled with n_h copies of the newest observed frame:

    for every g ≠ g*, for every p:   v_g[p] = x_g[n_h]           (computed once, reused)

The target window holds **physical** frames and starts from the observations:

    U_0 = ( x_{g*}[1] , x_{g*}[2] , … , x_{g*}[n_h] )

    for k = 1 … n_f:

        (i)   encode the current windows, all groups, every position:
              z^{(k)}[p] = cat_g ( E_g(v_g[p]) for g ≠ g* ,  E_{g*}( U_{k-1}[p] ) )

        (ii)  one processor step:
              ζ_k = Φ( cat_p( z^{(k)}[1] , … , z^{(k)}[n_h] ) )

        (iii) decode to a raw increment:
              r_k = D( ζ_k )                                       r_k ∈ ℝ^{1×H×W}

        (iv)  the anchor is the newest frame of the CURRENT target window, channel j only:
              a_k = U_{k-1}[n_h] [ j : j+1 ]

        (v)   the forecast for this lead:
              ŷ_k = F( r_k , a_k )

        (vi)  build the frame to append: a copy of U_{k-1}[n_h] in which channel j is
              replaced by ŷ_k and every other channel is left untouched (so the non-ice
              channels remain at their last observed values):
              u_k[ j ]  = ŷ_k
              u_k[ i ]  = U_{k-1}[n_h][ i ]      for all channels i ≠ j

        (vii) evolve the window:
              U_k = ( U_{k-1}[2] , … , U_{k-1}[n_h] , u_k )

Substituting (iv) into (v) and using (vi)–(vii), the target channel obeys the closed
recursion

    ŷ_k = M( clamp( ŷ_{k-1} + r_k , 0 , 1 ) ) ,      with the definition  ŷ_0 := x_{g*}[n_h][ j : j+1 ]

**A note for future investigation and improvements: non-target groups hold their last observed frame
throughout the rollout, and no lead index is passed to the processor, does this tie to the missing data issue as well maybe?**

**An important ansatz behind this is**: with physical rollout, encoder has a consistent role during the training which is to be trained model q( z_i | u_i ), with u_i be existing observation in physical space. with the caveat that when we go to forecasting steps of i, u_i has only the SIC channel updated, other channels reuse the last observation (at timestep n_h). That i believe effectively should act like constraint. Where as latent rollout allows z_i to evolve free without the encoder imposed constraint, and the encoder only has that role at steps before i>n_h, it encodes once at the beginning using all historical u_i. at forecasting range, the gradient passed back to update the encoder through a cascading composition map that is not soly interpretable as physical data space bounding but rather involved with the evolution itself. At current stage this is only an ansatz but these early test results seemed to show some promise but I am going to proceed with caution, a definitive test designed to just test this ansatz would be helpful 

### What is added:

Two opt-in switches on `EncodeProcessDecode`, both off by default; with neither set the
model is byte-identical to `main`.

| config | behaviour |
|---|---|
| *(nothing set)* | current `main`, unchanged |
| `skip_connection: additive` | #405 as merged — every lead corrects the newest observation |
| `rollout_space: physical` + `predict_residual: true` + `skip_connection: additive` | the loop is closed in observation space; each lead corrects the previous lead |

### Relationship to #405

Both formulations apply the same operator F, and F is #405's additive skip connection. The
residual tendency head of this PR is therefore not a second mechanism: it is #405's skip
connection evaluated against an anchor that advances with the lead instead of staying at the
forecast origin.

- **E_g** — unchanged, but applied n_h·n_f times instead of n_h times.
- **Φ** — unchanged, and called once per lead in both.
- **D** — unchanged; its output is read as an absolute field in §2.2 and as an increment r_k
  in §2.3.
- **F, R, M** — unchanged.
- **the anchor** — changed: fixed a in §2.2, advancing a_k in §2.3.
- **W_k** — absent from §2.3. The latents still exist there, recomputed by step (i) at every
  lead and then discarded; U_k is a window of physical frames, not a collection of ζ's.

The physical path now calls `finalise()` directly instead of re-implementing clamp-and-mask.
That refactor was **verified bitwise identical** to the previous hand-written version on a
residual model with a perturbed tendency head (maximum absolute difference 0.0).

Because the anchor is applied by the skip connection, `predict_residual: true` now requires
`skip_connection.method: additive`, and raises otherwise: without a skip connection
`finalise()` would silently drop the anchor and return an absolute prediction.

## Purpose of the updated anchor strategy

Persistence becomes directly representable. At lead 1 neither architecture has fed anything back, both are one encode → process → decode. So lead 1 isolates the anchor. Without an anchor the model cannot reproduce
its own input without some blurs: reconstructing the field requires passing it through a downsampling CNN, a
patch-embedding ViT bottleneck and an upsampling CNN. With an anchor, a zero tendency is
persistence exactly, so the model starts on that baseline and spends its capacity on the
change. `zero_init_tendency` (default true) zeroes the decoder's final convolution so the
first forward pass reproduces the anchor bit-exactly. Copying the input (persistence) is the reference:

| arm | val, lead-1 MAE vs copying | test, lead-1 MAE vs copying |
|---|---|---|
| no anchor (upstream) | **+22.4% worse** | **+23.4% worse** |
| anchored (this PR) | −0.7% (beats) | −0.4% (beats) |

## Compared on real data 

Real data `full_south`, lr 6e-4, no EMA, 40 epochs, seed 123, 336 validation / 349 test windows using the default split,
active-cell metrics

# Error tables — test year 2019 (349 windows, never used for selection)

Antarctic `full_south`, target `sic-ssmis`, 25 km EASE2 432x432, 3 history days -> 14 forecast days.
Each arm at its best-validation checkpoint. Computed by direct inference from saved checkpoints.

## 1. All metrics, active cells

Ratios are model / persistence averaged over all 14 leads — **lower is better, below 1.0 beats persistence**.
Sharpness = mean|grad(pred)| / mean|grad(truth)|, **1.0 is ideal**. IIEE (km2), DIIEE (km), centroid (km): lower better.
FSS, SSIM, accuracy (%): higher better. IIEE/DIIEE/FSS/SSIM/accuracy from PR #396 metric code, run verbatim.

| model | MAE ratio | RMSE ratio | extent ratio | sharpness | IIEE | DIIEE | FSS | SSIM | accuracy % | centroid |
|---|---|---|---|---|---|---|---|---|---|---|
| physical+anchor, patch 12, no argo | 0.8541 | 0.8272 | 0.2900 | 0.9748 | 723,040 | 19.6 | 0.6302 | 0.9695 | 99.38 | 2.0 |
| physical+anchor, patch 12, dropout 0.2 | 0.8631 | 0.8106 | 0.3411 | 0.9886 | 733,232 | 19.7 | 0.6303 | 0.9676 | 99.37 | 2.1 |
| physical+anchor, patch 12 | 0.8653 | 0.8331 | 0.2576 | 0.9768 | 745,101 | 20.1 | 0.6277 | 0.9687 | 99.36 | 2.2 |
| physical+anchor, patch 8 | 0.8659 | 0.7962 | 0.3471 | 0.9763 | 722,542 | 19.4 | 0.6315 | 0.9685 | 99.38 | 2.1 |
| physical+anchor, patch 12, seed 231 | 0.8660 | 0.8194 | 0.3838 | 0.9902 | 751,854 | 20.2 | 0.6294 | 0.9669 | 99.36 | 2.1 |
| physical+anchor, patch 24 | 0.8685 | 0.8209 | 0.2837 | 0.9929 | 724,584 | 19.5 | 0.6288 | 0.9681 | 99.38 | 2.0 |
| physical+anchor, patch 12, 15-epoch | 0.8769 | 0.8383 | 0.3566 | 0.9862 | 762,541 | 20.5 | 0.6265 | 0.9675 | 99.35 | 2.1 |
| physical+anchor, patch 12, deep | 0.8812 | 0.8390 | 0.2630 | 0.9763 | 759,394 | 20.5 | 0.6262 | 0.9684 | 99.35 | 2.0 |
| BASELINE: main architecture (latent rollout, no anchor) | 0.8829 | 0.8381 | 0.4036 | 0.8933 | 754,234 | 20.6 | 0.6161 | 0.9693 | 99.35 | 1.9 |

Per-metric winners differ: `patch 8` takes RMSE, IIEE, DIIEE, FSS and accuracy; `no argo` takes MAE and SSIM;
`patch 12` takes extent; the baseline takes centroid error. Mean rank across all ten: patch 8 and no argo tie at 3.20, baseline 6.90.

optimal-epoch (early stop) vs all-epoch (MAE ratio, test 2019)
| model | optimal ckpt | end-of-training | degradation |
|---|---|---|---|
| BASELINE: latent rollout, no anchor | 0.8829 (ep 19) | 0.8893 (ep 39) | **+0.7%** |
| physical+anchor, patch 12, no argo | 0.8541 (ep 9) | 0.8804 | +3.1% |
| physical+anchor, patch 12, seed 231 | 0.8660 (ep 2) | 0.9015 | +4.1% |
| physical+anchor, patch 12, dropout 0.2 | 0.8631 (ep 2) | 0.9032 | +4.6% |
| physical+anchor, patch 8 | 0.8659 (ep 2) | 0.9065 | +4.7% |
| physical+anchor, patch 12, deep | 0.8812 (ep 9) | 0.9233 | +4.8% |
| physical+anchor, patch 12 | 0.8653 (ep 9) | 0.9104 | +5.2% |
| physical+anchor, patch 24 | 0.8685 (ep 4) | 0.9351 | +7.7% |

## 2. Active cells vs whole grid (whole grid is the upstream convention, included here for reference)

Same forecasts; the columns differ only in which cells are averaged. The team's metrics apply no mask.
MAE_all = 0.265813 x MAE_active and RMSE_all = 0.515570 x RMSE_active exactly, because prediction, truth
and persistence are all identically zero outside the mask. Accuracy is the exception and genuinely differs.

| model | MAE L1 active | MAE L1 whole | MAE L14 active | MAE L14 whole | RMSE L14 active | RMSE L14 whole | acc % L14 active | acc % L14 whole |
|---|---|---|---|---|---|---|---|---|
| BASELINE: main architecture | 0.0164 | 0.0044 | 0.0424 | 0.0113 | 0.1111 | 0.0573 | 96.3156 | 99.0206 |
| OURS: patch 8 | 0.0129 | 0.0034 | 0.0430 | 0.0114 | 0.1038 | 0.0535 | 96.4378 | 99.0531 |
| OURS: no argo | 0.0125 | 0.0033 | 0.0425 | 0.0113 | 0.1112 | 0.0573 | 96.3815 | 99.0382 |
| patch 12 | 0.0128 | 0.0034 | 0.0429 | 0.0114 | 0.1121 | 0.0578 | 96.2709 | 99.0088 |
| patch 24 | 0.0128 | 0.0034 | 0.0434 | 0.0115 | 0.1090 | 0.0562 | 96.4101 | 99.0457 |
| dropout 0.2 | 0.0128 | 0.0034 | 0.0428 | 0.0114 | 0.1064 | 0.0548 | 96.3280 | 99.0239 |
| **PERSISTENCE** | **0.0128** | **0.0034** | **0.0566** | **0.0151** | **0.1505** | **0.0776** | **94.7215** | **98.5969** |

At lead 14 the model-vs-persistence accuracy margin is **1.68 points** on active cells ( **0.44** on the whole grid).

fig08_sharpness_final
fig06_extent_final
fig02_mae_ratio_final
fig13_mae_wholegrid

video_test2019_lead1
video_test2019_lead14

### DIFF
--- a/icenet_mp/models/encode_process_decode.py
+++ b/icenet_mp/models/encode_process_decode.py
@@ -40,6 +40,10 @@ class EncodeProcessDecode(BaseModel):
             encoders: config for the per-input-group encoders (plus ``latent_space``).
             processor: config for the latent-space processor.
             decoder: config for the decoder producing the output space.
+            target_variable_indices: indices, within the target INPUT group, of the
+                variable(s) being predicted. Must select exactly as many channels as
+                ``output_space`` has; used to pull the newest observed target frame
+                out of the input window as the skip connection's anchor.
             mask_dir: directory holding the mask ``.npy`` files, if masking is used.
             **kwargs: forwarded to ``BaseModel`` (spaces, steps, optimiser, loss, ...).
 
@@ -105,48 +109,10 @@ class EncodeProcessDecode(BaseModel):
             raise ValueError(msg)
         self.target_variable_indices = target_variable_indices
 
-        if rollout_space not in {"latent", "physical"}:
-            msg = (
-                f"rollout_space must be 'latent' or 'physical', got {rollout_space!r}."
-            )
-            raise ValueError(msg)
+        self._validate_rollout_options(rollout_space, predict_residual, decoder)
         self.rollout_space = rollout_space
         self.predict_residual = bool(predict_residual)
         self.feedback_channel = feedback_channel
-        if self.predict_residual and rollout_space != "physical":
-            msg = (
-                "predict_residual=True requires rollout_space='physical': the residual "
-                "is added to the previous PHYSICAL field, which only exists as a "
-                "rollout state in physical space."
-            )
-            raise ValueError(msg)
-        # The residual update is applied by the decoder's additive skip connection
-        # (#405), must be configured: finalise() adds the anchor only when
-        # skip_connection is present, and would otherwise silently drop it and
-        # return an absolute prediction (rather than the tendency).
-        skip_method = str(
-            (decoder.get("skip_connection") or {}).get("method", "none")
-        )
-        if self.predict_residual and skip_method != "additive":
-            msg = (
-                f"predict_residual=True requires the decoder to use an additive skip "
-                f"connection (got skip_connection.method={skip_method!r}): the "
-                f"tendency is added to the anchor by the decoder's skip connection, "
-                f"so without it the anchor would be dropped and the output would be "
-                f"an absolute prediction rather than a residual one."
-            )
-            raise ValueError(msg)
-        if (
-            self.predict_residual
-            and str(decoder.get("restrict_range") or "none") != "none"
-        ):
-            msg = (
-                f"predict_residual=True requires the decoder's restrict_range to be "
-                f"'none' (got {decoder.get('restrict_range')!r}): the decoder emits a "
-                f"signed tendency, which must not be squashed into [0, 1]. The sum "
-                f"previous + tendency is clamped to [0, 1] by the model instead."
-            )
-            raise ValueError(msg)
 
         # Add one encoder per dataset
         # We store this as a list to ensure consistent ordering
@@ -230,6 +196,57 @@ class EncodeProcessDecode(BaseModel):
         # forward pass reproduces the last observation exactly at every lead.
         if self.predict_residual and zero_init_tendency:
             self._zero_tendency_head()
+
+    @staticmethod
+    def _validate_rollout_options(
+        rollout_space: str,
+        predict_residual: bool,  # noqa: FBT001 - mirrors the __init__ keyword
+        decoder: DictConfig,
+    ) -> None:
+        """Reject rollout/residual settings that cannot work, before anything is built.
+
+        Kept out of `__init__` so that adding a check does not push it past the
+        cyclomatic-complexity limit.
+        """
+        if rollout_space not in {"latent", "physical"}:
+            msg = (
+                f"rollout_space must be 'latent' or 'physical', got {rollout_space!r}."
+            )
+            raise ValueError(msg)
+        if not predict_residual:
+            return
+
+        if rollout_space != "physical":
+            msg = (
+                "predict_residual=True requires rollout_space='physical': the residual "
+                "is added to the previous PHYSICAL field, which only exists as a "
+                "rollout state in physical space."
+            )
+            raise ValueError(msg)
+
+        # The residual update is applied by the decoder's additive skip connection
+        # (#405), so one must be configured: finalise() adds the anchor only when
+        # skip_connection is present, and would otherwise silently drop it and
+        # return an absolute prediction (rather than the tendency).
+        skip_method = str((decoder.get("skip_connection") or {}).get("method", "none"))
+        if skip_method != "additive":
+            msg = (
+                f"predict_residual=True requires the decoder to use an additive skip "
+                f"connection (got skip_connection.method={skip_method!r}): the "
+                f"tendency is added to the anchor by the decoder's skip connection, "
+                f"so without it the anchor would be dropped and the output would be "
+                f"an absolute prediction rather than a residual one."
+            )
+            raise ValueError(msg)
+
+        if str(decoder.get("restrict_range") or "none") != "none":
+            msg = (
+                f"predict_residual=True requires the decoder's restrict_range to be "
+                f"'none' (got {decoder.get('restrict_range')!r}): the decoder emits a "
+                f"signed tendency, which must not be squashed into [0, 1]. The sum "
+                f"previous + tendency is clamped to [0, 1] by the model instead."
+            )
+            raise ValueError(msg)
 
     def _zero_tendency_head(self) -> None:
         """Zero the decoder's output convolution so the initial tendency is exactly 0."""
@@ -352,7 +369,7 @@ class EncodeProcessDecode(BaseModel):
             raw = self.decoder(step_latent)
 
             if self.predict_residual:
-                # Compared to the latent path the difference here is 
+                # Compared to the latent path the difference here is
                 # the anchor: here we use the state produced by the previous forecast step.
                 anchor = self._anchor(target_window, n_out)
                 field = self.decoder.finalise(raw, anchor)
--- a/icenet_mp/models/encode_process_decode.py
+++ b/icenet_mp/models/encode_process_decode.py
@@ -44,29 +44,25 @@ class EncodeProcessDecode(BaseModel):
             **kwargs: forwarded to ``BaseModel`` (spaces, steps, optimiser, loss, ...).
 
         Args:
-            rollout_space: "latent" (default, unchanged behaviour) or "physical".
+            rollout_space: "latent" (default, unchanged behaviour) or "physical" (residual learning, autoregressive-like).
 
-                "latent" is the historical path: the processor predicts the next
-                COMBINED LATENT and that latent is appended to its own input window
-                (`BaseProcessor.rollout`). Two consequences, both measured as real
-                failures on 2026-07-29: (a) nothing constrains a processor-produced
+                "latent" option is the original method: the processor predicts the next
+                combined latent, which is then appended to its own input window
+                (BaseProcessor.rollout). Potentiallly two flagged consequences, from tested
+                failures (eg, on 2026-07-29): (1) no constrains on a processor to produce a
                 latent to lie in the distribution the encoders produce, and there is no
-                re-encoding or consistency term, so from forecast step 2 onward the
-                window is fed vectors from a different distribution; (b) the window
-                carries EVERY input group's latent channels, but only the target
-                group's are supervised (the loss sees the decoded target alone), so a
-                multi-input model invents unsupervised ERA5/Argo latents and then
-                conditions on them. The optimiser's cheapest defence against both is a
-                near-idempotent map, which is exactly the near-fixed-point (static
-                forecast) behaviour observed.
+                re-encoding (consistency term), so from forecast step 2 onward the
+                window is fed vectors from a different distribution; (2) the roll-out window
+                takes every input group's latent channels and carry them forward in time, so a
+                multi-input model forecasts uncontraint ERA5/Argo latents and then
+                conditions on them. The optimiser tends to find the easiest "optimum" to satisfying both, which is a near-idempotent map, which likely explains the almost fixed-point (static forecast) behaviour observed.
 
-                "physical" closes the loop in observation space instead: encode the
-                window of PHYSICAL frames, take one processor step, decode to a
-                physical field, roll that field into the window, and re-encode. The
-                fed-back state is therefore always something the encoders were trained
-                on. Groups other than the target hold their last observed frame (we
-                have no atmospheric forecast; persisting is an assumption we state
-                rather than a latent we hallucinate).
+                "physical" option wraps the loop in observation space instead: encode the
+                window of physical frames, take one processor step (roll-out and forecast),
+                decode back to a physical field, roll that field into the window, and re-encode.
+                The feedback state in this option is always something the encoders were trained
+                on. Groups other than the SIC (target) keep their last observed frame (we
+                dont have atmospheric forecast; persisting is an assumption and we can maybe improve this, but it should prevent a hallucinated latent that might get trapped on persistent latent for the reason given above).
 
             predict_residual: if True the decoder output is a TENDENCY added to the
                 previous field, so `prediction = clamp(previous + delta, 0, 1)`.
@@ -74,27 +70,25 @@ class EncodeProcessDecode(BaseModel):
                 This is the fix for the binding failure: with the absolute
                 parameterisation, reproducing the input requires pushing it through a
                 downsampling CNN, a patch-embedding ViT bottleneck and an upsampling
-                CNN, so PERSISTENCE IS NOT IN THE MODEL'S HYPOTHESIS SPACE — measured
+                CNN, so persistence is not in the model's hypothesis space. Measured
                 on the toy, the model scores active-cell MAE 0.093 at lead 1 where
                 merely copying the input scores 0.034. With a residual head, delta = 0
-                IS persistence, exactly and for free, and every parameter is spent on
-                the change instead of on rebuilding the state. Requires the decoder's
-                `restrict_range` to be "none": the bounding is applied to the SUM here,
+                is exactly the persistence, and every parameter is trained to predict
+                the change instead of rebuilding the persistence state. This Requires decoder's
+                restrict_range to be set to "none": the bounding is applied to the sum here,
                 not to the increment.
 
-            zero_init_tendency: with `predict_residual=True`, zero the decoder's final
-                convolution at initialisation so the model STARTS EXACTLY AT
-                PERSISTENCE and training can only move away from it deliberately.
-                Without this, a randomly initialised tendency head emits a large signed
+            zero_init_tendency: with predict_residual=True, zero the decoder final
+                convolution at initialisation so the model starts at persistence exactly,
+                and training can move away from it only to reduce the loss (intended).
+                Otherwise a random initialisation of the tendency produces a large signed
                 field, the sum saturates against the [0, 1] clamp, and the first epochs
-                are spent climbing back down to a sane state — the regime in which every
-                run so far ended up worse than persistence. Ignored when
-                `predict_residual=False`.
+                are spent climbing back down to sanity state (this is also observed in real training runs), a mechanism that might be partially responsible for many previous runs ended up worse than persistence.
 
-            feedback_channel: index, within the target input group, of the channel that
-                the prediction overwrites when it is rolled back into the window under
-                "physical" rollout. Only needed when the target group has more channels
-                than the model outputs (production `sic-ssmis` has 6, the output has 1);
+            feedback_channel: index of the channel within the target input group (SIC) that
+                the prediction overwrites when it is rolled back into the window under the
+                physical rollout option. This is only needed when the target group has more channels
+                than the model outputs (eg, production sic-ssmis has 6, the output has 1);
                 when the counts match, all channels are replaced.
 
         """
@@ -124,6 +118,22 @@ class EncodeProcessDecode(BaseModel):
                 "predict_residual=True requires rollout_space='physical': the residual "
                 "is added to the previous PHYSICAL field, which only exists as a "
                 "rollout state in physical space."
+            )
+            raise ValueError(msg)
+        # The residual update is applied by the decoder's additive skip connection
+        # (#405), must be configured: finalise() adds the anchor only when
+        # skip_connection is present, and would otherwise silently drop it and
+        # return an absolute prediction (rather than the tendency).
+        skip_method = str(
+            (decoder.get("skip_connection") or {}).get("method", "none")
+        )
+        if self.predict_residual and skip_method != "additive":
+            msg = (
+                f"predict_residual=True requires the decoder to use an additive skip "
+                f"connection (got skip_connection.method={skip_method!r}): the "
+                f"tendency is added to the anchor by the decoder's skip connection, "
+                f"so without it the anchor would be dropped and the output would be "
+                f"an absolute prediction rather than a residual one."
             )
             raise ValueError(msg)
         if (
@@ -216,8 +226,8 @@ class EncodeProcessDecode(BaseModel):
             mask_dir=mask_dir,
         )
 
-        # Start the trajectory AT persistence: a zeroed tendency head means the model's
-        # first forward pass reproduces the last observation exactly at every lead.
+        # Start the trajectory at exactly the persistence: a zeroed tendency head means the first
+        # forward pass reproduces the last observation exactly at every lead.
         if self.predict_residual and zero_init_tendency:
             self._zero_tendency_head()
 
@@ -226,9 +236,9 @@ class EncodeProcessDecode(BaseModel):
         final = cast("torch.nn.Sequential", self.decoder.model)[-1]
         if not isinstance(final, torch.nn.Conv2d):
             msg = (
-                f"zero_init_tendency=True expects the decoder to end in a Conv2d so the "
-                f"tendency can be zeroed, but {type(self.decoder).__name__} ends in "
-                f"{type(final).__name__}. Pass zero_init_tendency=false to skip this."
+                f"zero_init_tendency=True requires the decoder to end in a Conv2d so the "
+                f"tendency can be 0, but {type(self.decoder).__name__} ends in "
+                f"{type(final).__name__}. Try pass zero_init_tendency=false to skip this."
             )
             raise TypeError(msg)
         with torch.no_grad():
@@ -279,7 +289,7 @@ class EncodeProcessDecode(BaseModel):
         return self.decoder.rollout(latent_output, persistence)
 
     def _forward_physical(self, inputs: dict[str, TensorNTCHW]) -> TensorNTCHW:
-        """Autoregressive rollout closed in PHYSICAL space, one forecast step at a time.
+        """Autoregressive-like rollout closed in physical/observation space one forecast step at a time.
 
         Per step: encode the window of observed/predicted frames, take ONE processor
         step, decode to a physical field, then roll that field into the window and
@@ -342,12 +352,13 @@ class EncodeProcessDecode(BaseModel):
             raw = self.decoder(step_latent)
 
             if self.predict_residual:
+                # Compared to the latent path the difference here is 
+                # the anchor: here we use the state produced by the previous forecast step.
                 anchor = self._anchor(target_window, n_out)
-                field = self.decoder.mask((anchor + raw).clamp(0.0, 1.0))
+                field = self.decoder.finalise(raw, anchor)
             else:
-                # main's finalise() now takes the skip/anchor frame as a second,
-                # required argument (#405). The non-residual physical path has no
-                # anchor of its own, so pass None -- the decoder then applies only
+                # The non-residual physical path has no
+                # anchor of its own, so pass None; the decoder then applies only
                 # range restriction and masking, exactly as before.
                 field = self.decoder.finalise(raw, None)
             outputs.append(field)
--- a/icenet_mp/models/encode_process_decode.py
+++ b/icenet_mp/models/encode_process_decode.py
@@ -1,10 +1,10 @@
-from typing import TYPE_CHECKING, Any, ClassVar
+from typing import TYPE_CHECKING, Any, ClassVar, cast
 
 import hydra
 import torch
 from omegaconf import DictConfig
 
-from icenet_mp.types import DataSpace, TensorNTCHW
+from icenet_mp.types import DataSpace, TensorNCHW, TensorNTCHW
 
 from .base_model import BaseModel
 
@@ -20,7 +20,7 @@ class EncodeProcessDecode(BaseModel):
     # Parameters that should be excluded from hyperparameter logging (e.g. local paths)
     ignored_hparams: ClassVar[frozenset[str]] = BaseModel.ignored_hparams | {"mask_dir"}
 
-    def __init__(
+    def __init__(  # noqa: PLR0913 - config-driven keywords, all defaulted
         self,
         *,
         encoders: DictConfig,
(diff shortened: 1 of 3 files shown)


jemrobinson:
@Shaerdan: are you refactoring this to modify the skip-connections added in #405 or do you want me to pick it up?

Shaerdan:
> @Shaerdan: are you refactoring this to modify the skip-connections added in #405 or do you want me to pick it up?

@jemrobinson I will take care of this and send a review request to you after modifying #405

Shaerdan:
i am now refactoring this PR to modify PR405.

icenet_mp/models/encode_process_decode.py
```diff
@@ -126,6 +192,77 @@ def __init__(
...
+        if str(decoder.get("restrict_range") or "none") != "none":
+            msg = (
+                f"predict_residual=True requires the decoder's restrict_range to be "
+                f"'none' (got {decoder.get('restrict_range')!r}): the decoder emits a "
+                f"signed tendency, which must not be squashed into [0, 1]. The sum "
+                f"previous + tendency is clamped to [0, 1] by the model instead."
+            )
+            raise ValueError(msg)
+
+    def _zero_tendency_head(self) -> None:
```
jemrobinson, line 251: Could/should this be a config argument to the decoder instead of to EncodeProcessDecode?

icenet_mp/models/encode_process_decode.py
```diff
@@ -126,6 +192,77 @@ def __init__(
...
+                f"an absolute prediction rather than a residual one."
+            )
+            raise ValueError(msg)
+
+        if str(decoder.get("restrict_range") or "none") != "none":
+            msg = (
+                f"predict_residual=True requires the decoder's restrict_range to be "
+                f"'none' (got {decoder.get('restrict_range')!r}): the decoder emits a "
+                f"signed tendency, which must not be squashed into [0, 1]. The sum "
+                f"previous + tendency is clamped to [0, 1] by the model instead."
```
jemrobinson, line 247: I think you might not need this. See https://github.com/alan-turing-institute/icenet-mp/blob/3b315366273d2607c24d8a6e5b2c2015ba7579dc/icenet_mp/models/decoders/base_decoder.py#L62-L73 which (a) sets the range of the RestrictRange layer to be symmetric around 0 and (b) allows `restrict_range_min` and `restrict_range_max` to be specified in the config.

Shaerdan: thanks for pointing that out, i am behind that commit. Will update it

Shaerdan:
i want to do more tests on this PR to make sure i have a clear view on what exactly this rollout strategy brings: on the toy i observed clear improvements and double checked, confident that the results there are genuine; on the full_south real dataset, the advantage disappeared almost, there are still some improvements but not on temporal tracking. do you recommend i refine the codebase and we go with the review already @jemrobinson  or wait for few days so I can do more tests and gather more evidence?

jemrobinson:
Happy for you to do some more tests. If you want a review at any point, just let me know.

Shaerdan:
hi @jemrobinson, I finished a summary from recent test results, and handing this over to you for your review, thanks!

icenet_mp/models/encode_process_decode.py
```diff
@@ -20,17 +20,82 @@ class EncodeProcessDecode(BaseModel):
...
         *,
         encoders: DictConfig,
         processor: DictConfig,
         decoder: DictConfig,
         target_variable_indices: list[int],
         mask_dir: str | None = None,
+        rollout_space: str = "latent",
+        predict_residual: bool = False,
+        feedback_channel: int | None = None,
+        zero_init_tendency: bool = True,
```
jemrobinson, line 34: `zero_init_tendency` should be an argument to the decoder, not to the EncodeProcessDecode model. Could be either added to BaseDecoder (and only used if appropriate) or added independently to each decoder that has a final convolutional layer.

icenet_mp/models/encode_process_decode.py
```diff
@@ -20,17 +20,82 @@ class EncodeProcessDecode(BaseModel):
...
         self,
         *,
         encoders: DictConfig,
         processor: DictConfig,
         decoder: DictConfig,
         target_variable_indices: list[int],
         mask_dir: str | None = None,
+        rollout_space: str = "latent",
+        predict_residual: bool = False,
+        feedback_channel: int | None = None,
```
jemrobinson, line 33: The anchoring logic (keep track of which channels in the latent space correspond to SIC input and only overwrite those) is also done in the DDPMProcessor. This could probably be delegated to the processor.

I think it is also potentially fragile as we might, for example, decide that the best way to combine input datasets in the latent space is via a convolution (or similar) rather than just `torch.cat`.

icenet_mp/models/encode_process_decode.py
```diff
@@ -20,17 +20,82 @@ class EncodeProcessDecode(BaseModel):
...
+                out of the input window as the skip connection's anchor.
+            mask_dir: directory holding the mask ``.npy`` files, if masking is used.
+            **kwargs: forwarded to ``BaseModel`` (spaces, steps, optimiser, loss, ...).
+
+        Args:
+            rollout_space: "latent" (default, unchanged behaviour) or "physical" (residual learning, autoregressive-like).
+
+                "latent" option is the original method: the processor predicts the next
+                combined latent, which is then appended to its own input window
+                (BaseProcessor.rollout). Potentiallly two flagged consequences, from tested
```
jemrobinson, line 55: A lot of this text refers to unspecified issues that are fixed by these changes. I think this should go into the PR, and the text here should describe the state we'll be in once this is merged. For example "this means that predicted ERA5/Argo latents are replaced with persistence at each step of rollout".

jemrobinson (changes requested):
Sorry - just noticed I hadn't pushed the button on this review!

icenet_mp/models/encode_process_decode.py
Shaerdan: I agree, am working on making the change

icenet_mp/models/encode_process_decode.py
Shaerdan: agreed, i am moving this out,

icenet_mp/models/encode_process_decode.py
Shaerdan: yes, I'm on it

icenet_mp/models/encode_process_decode.py
```diff
@@ -135,7 +272,13 @@ def forward(self, inputs: dict[str, TensorNTCHW]) -> TensorNTCHW:
         - process in latent space [NTCHW] [batch, n_forecast_steps, n_latent_channels_total, H_latent, W_latent]
         - decode back to [NTCHW] output space [batch, n_forecast_steps, n_output_channels, H_output, W_output]
         - add a skip connection from the most recent target value to every forecast step
+
+        When `rollout_space="physical"` the loop is closed in observation space instead;
+        see `_forward_physical`.
         """
+        if self.rollout_space == "physical":
```
sylvesterkaczmarek, line 279: One thing to watch when rebasing this onto current `main`: `EncodeProcessDecode.training_step()` now performs `encode_inputs -> processor.rollout -> decoder.rollout` directly rather than calling `forward()`. Since the physical-rollout dispatch currently lives here in `forward()`, wouldn't training still use the latent rollout while inference uses `_forward_physical()`?

If so, I think the physical path also needs to be integrated into `training_step()` (or factored into a shared method used by both), otherwise the model would be trained and evaluated with different rollout semantics.

icenet_mp/models/encode_process_decode.py
```diff
@@ -160,7 +303,101 @@ def forward(self, inputs: dict[str, TensorNTCHW]) -> TensorNTCHW:
...
+            target_window = self._advance(target_window, field)
+
+        return torch.stack(outputs, dim=1)
+
+    def _anchor(self, target_window: TensorNTCHW, n_out: int) -> TensorNCHW:
+        """Return the current physical state that a residual is added to."""
+        newest = target_window[:, -1]  # (B, C_t, H, W)
+        if self.feedback_channel is None:
+            return newest
+        idx = int(self.feedback_channel)
```
sylvesterkaczmarek, line 392: Could this reuse `target_variable_indices` rather than introducing a separate `feedback_channel`?

`target_variable_indices` already defines which channels of the target input correspond to the model outputs. With the current implementation the two settings can disagree, and `idx : idx + n_out` also assumes that multiple predicted variables are contiguous.

Using `target_variable_indices` for both `_anchor()` and `_advance()` would seem to keep the physical feedback consistent with the existing target definition and also support non-contiguous target variables.

sylvesterkaczmarek:
@jemrobinson, @Shaerdan one question about the real-data comparison: the baseline here is `latent rollout, no anchor`, while the proposed models are `physical + anchor`.

Since #405 already showed that the additive anchor/skip connection itself can provide a substantial improvement, wouldn't it be useful to include a matched `latent rollout + additive anchor` arm with the same architecture, inputs, seed and training setup?

That would isolate the effect of closing the loop in physical space from the effect of adding the anchor, and make it much clearer what improvement is specifically attributable to the rollout strategy.

sylvesterkaczmarek:
Also, a question about the treatment of non-target groups in the physical rollout. As written, `v_g[p] = x_g[n_h]` for every history position means that even at lead 1 the observed ERA5/Argo history is replaced by `n_h` copies of the newest frame.

Would it be preferable to initialise those windows with their real observed history, and only append/persist the newest observation as the forecast window advances? For example, lead 1 would see `(x1, x2, x3)`, lead 2 `(x2, x3, x3)`, then `(x3, x3, x3)`.

That would preserve all available historical information at forecast origin while still avoiding forecasting unsupervised non-target variables.

Shaerdan:
> Also, a question about the treatment of non-target groups in the physical rollout. As written, `v_g[p] = x_g[n_h]` for every history position means that even at lead 1 the observed ERA5/Argo history is replaced by `n_h` copies of the newest frame.
> 
> Would it be preferable to initialise those windows with their real observed history, and only append/persist the newest observation as the forecast window advances? For example, lead 1 would see `(x1, x2, x3)`, lead 2 `(x2, x3, x3)`, then `(x3, x3, x3)`.
> 
> That would preserve all available historical information at forecast origin while still avoiding forecasting unsupervised non-target variables.

thanks for raising this point. I meant to denote lead 1 as the day we already dont have any observations (i followed 1 based indexing in the algorithm description part, where 1 to n_f are all in the forecast group), unless i mistaken the data availability, i believe lead 1 is where SIC data is not available and I assumed ERA5 etc also would be, but i should look into this cautiously to make sure. In either case yes the idea is to only roll SIC forward and keep ERA5 and other predictors as the data until they are no longer available.

Shaerdan:
sorry, I misclicked "close with comment" button, reopening this

icenet_mp/models/encode_process_decode.py
Shaerdan: rewritten as a short and describe only docstring, aligned with the rest of the codebase

icenet_mp/models/encode_process_decode.py
Shaerdan: Removed and the old rejection test now pins the opposite (a bounded residual decoder
is accepted).

icenet_mp/models/encode_process_decode.py
Shaerdan: Moved as suggested

icenet_mp/models/encode_process_decode.py
Shaerdan: Thanks, yes they do look alike, but they address two different needs.

The DDPM processor addresses "where is the target after the encoders have run?", the answer is a position inside the combined latent tensor (target_channel_offset), which relies on how many latent channels each encoder produces and on the order they are concatenated. If we ever fused the encoders' outputs differently (a convolution instead of torch.cat), that position would stop meaning anything.

The physical rollout asks "where is the target before any encoder runs?" and the answer here is a position inside the raw sea-ice input frame, ie, channel ice_conc out of the six channels the dataset provides. That position comes from the variable names in the data and does not change however the latents are fused.

Following @sylvesterkaczmarek's suggestion below, the physical rollout no longer keeps its own copy of that position. It uses target_variable_indices, the list the model already receives from the data module and that get_persistence and Persistence class already use, so "which channels are the target" is defined in exactly one place.

Moving the write-back into the processor would mean processors have to know about encoders and decoders, which they currently don't. Making the DDPM side use the same list instead of an offset is what #440 asks for; I'd be very happy to do that as a follow-up, or co-PR it.

icenet_mp/models/encode_process_decode.py
Shaerdan: spot on, fixed in the merge with current main: training_step now routes rollout_space="physical" through forward(), and TestTrainEvalParity ensures that training and validation score the same path (and that the latent path is unchanged by the routing)

icenet_mp/models/encode_process_decode.py
Shaerdan: Done. `_anchor` and `_advance` now index by `target_variable_indices`; `feedback_channel` is removed from the model and from the config; a new test covers non-contiguous target variables (channels 0 and 2 of a 3-channel group). For the single variable case the output is bit identical to before.

Shaerdan:
> @jemrobinson, @Shaerdan one question about the real-data comparison: the baseline here is `latent rollout, no anchor`, while the proposed models are `physical + anchor`.
> 
> Since #405 already showed that the additive anchor/skip connection itself can provide a substantial improvement, wouldn't it be useful to include a matched `latent rollout + additive anchor` arm with the same architecture, inputs, seed and training setup?
> 
> That would isolate the effect of closing the loop in physical space from the effect of adding the anchor, and make it much clearer what improvement is specifically attributable to the rollout strategy.

Agreed. A matched arm with the latent rollout plus the additive anchor (same architecture, inputs, seed and recipe) is part of the planned comparison, precisely to separate the anchor from closing the loop in physical space. I'll post the numbers when those runs have landed but I think they should be in a different PR or maybe i can open an issue for that.

Shaerdan:
> > Also, a question about the treatment of non-target groups in the physical rollout. As written, `v_g[p] = x_g[n_h]` for every history position means that even at lead 1 the observed ERA5/Argo history is replaced by `n_h` copies of the newest frame.
> > Would it be preferable to initialise those windows with their real observed history, and only append/persist the newest observation as the forecast window advances? For example, lead 1 would see `(x1, x2, x3)`, lead 2 `(x2, x3, x3)`, then `(x3, x3, x3)`.
> > That would preserve all available historical information at forecast origin while still avoiding forecasting unsupervised non-target variables.
> 
> thanks for raising this point. I meant to denote lead 1 as the day we already dont have any observations (i followed 1 based indexing in the algorithm description part, where 1 to n_f are all in the forecast group), unless i mistaken the data availability, i believe lead 1 is where SIC data is not available and I assumed ERA5 etc also would be, but i should look into this cautiously to make sure. In either case yes the idea is to only roll SIC forward and keep ERA5 and other predictors as the data until they are no longer available.

Just realised i misread your comment here, ignore my previous answer pls. Yes I think it's worth try out what you suggested, sounds more reasonable than always fixing them to x3. that said, probably no harm to use x3 era5+argo since they only act as contraint via encoder to model q(z_i|x_i), the disconnection of time step later might well just push the model to disassociate time gap and try to train the encoder to learn "physical space constraint" so z_i wont go wild, was my original design ansats. would like to try this in a new PR as a refinement/improvement step (maybe i will call it fine-tuning physical rollout architecture).

sylvesterkaczmarek (approved):
Re-reviewed current `0a2dd8f`. The two concrete issues I raised are resolved: training and evaluation now use the same physical-rollout path, and physical feedback is keyed by `target_variable_indices`, including non-contiguous targets. The current test/lint matrix is green. The non-target observed-history treatment discussed above is reasonable as a follow-up experiment rather than a blocker here. No remaining blocker from me.

sylvesterkaczmarek (approved):
Re-reviewed current `c73cef4c`. The only delta since my prior approval is the residual config key update from `sic-icenet` to `sic-osisaf`, matching the renamed OSISAF group used by the full data configs. The test, lint, and documentation workflows are green. No remaining blocker from me.

sylvesterkaczmarek (approved):
Rechecked current `01ea2586` after the climatology merge from `main`. The PR-specific physical-rollout diff remains the same four files, including the target-index handling and matched train/eval rollout path I previously reviewed, and the Python 3.11-3.14 matrix is green. No new blocker from me.

jemrobinson:
👀 reviewing this

sylvesterkaczmarek (approved):
Re-reviewed current `4f1eeb02`. The latest two commits refactor `EncodeProcessDecode`/`ProcessorStage` construction and the zero-init check without changing the physical-rollout semantics I previously reviewed. The matched train/eval path and target-index handling remain intact, and the Python 3.11-3.14 matrix is green. No new blocker from me.

icenet_mp/models/encode_process_decode.py
```diff
@@ -185,6 +291,90 @@ def forward(self, inputs: dict[str, TensorNTCHW]) -> TensorNTCHW:
...
+                [encoder.rollout(windows[encoder.name]) for encoder in self.encoders],
+                dim=2,
+            )  # (B, n_history, C_latent_total, h, w)
+
+            # One processor step: the window is concatenated along channels, oldest to
+            # newest, exactly as BaseProcessor.rollout does it.
+            step_in = torch.cat(
+                [latent[:, idx_t] for idx_t in range(self.n_history_steps)], dim=1
+            )
+            step_latent = self.processor(step_in)
```
jemrobinson: This won't work for non-autoregressive `Processor`s (like `GSTAProcessor`) as they only implement `rollout` (NTCHW->NTCHW) and not `forward` (NCHW->NCHW).

I have some thoughts about how we could fix this - let's discuss later.

icenet_mp/models/encode_process_decode.py
```diff
@@ -185,6 +259,121 @@ def forward(self, inputs: dict[str, TensorNTCHW]) -> TensorNTCHW:
...
+        target_window = inputs[target_name].clone()  # (B, n_history, C_t, H_t, W_t)
+
+        outputs: list[TensorNCHW] = []
+        for _ in range(self.n_forecast_steps):
+            # Set target window to the most recent observation/prediction
+            windows[target_name] = target_window
+
+            # Encode inputs to latent space
+            # -> (B, n_history, C_latent_total, H_latent, W_latent)
+            latent: TensorNTCHW = self.encode_inputs(windows)
```
jemrobinson: This is a bit inefficient, since we're re-encoding the non-target variables at every step.

jemrobinson (changes requested):
I don't think this works for non-autoregressive Processors.

jemrobinson:
I ran a comparison of the latent rollout vs. physical rollout on the same dataset.

## SIC predictions

<table>
<tr>
<td>
media_images_output_static_2024-12-29-sea-ice-concentration_6_b107134a9e1159800ddd
</td>
<td>
media_images_output_static_2024-12-29-sea-ice-concentration_6_b9d05d47fdf90c5c01a0
</td>
</tr>
</table>

These look almost identical

## Metrics

W B Chart 23_09_2026, 14_16_44
W B Chart 23_09_2026, 14_17_33
W B Chart 23_09_2026, 14_17_39
W B Chart 23_09_2026, 14_17_43
W B Chart 23_09_2026, 14_17_51
W B Chart 23_09_2026, 14_17_56
W B Chart 23_09_2026, 14_18_09
W B Chart 23_09_2026, 14_18_15
W B Chart 23_09_2026, 14_18_18
W B Chart 23_09_2026, 14_18_23
W B Chart 23_09_2026, 14_18_26
W B Chart 23_09_2026, 14_18_35
W B Chart 23_09_2026, 14_18_38
W B Chart 23_09_2026, 14_18_42
W B Chart 23_09_2026, 14_18_46
W B Chart 23_09_2026, 14_18_50
W B Chart 23_09_2026, 14_18_53

These also look very similar.

**Conclusion:** rollout in physical space is valid, but doesn't seem to be adding much model improvement at the moment.

Are you happy to merge @Shaerdan ?

jemrobinson (approved):
LGTM. We agreed that it's fine for this path to explicit fail for non-autoregressive processors.

sylvesterkaczmarek:
Thanks all. Good to see the physical-rollout path land with the train/eval behavior aligned and the non-autoregressive case made explicit. The comparison results are also useful context for the next refinement rather than overclaiming improvement here.

PR #422, opened 2026-08-10
Add DDPM processor

### LINKED ISSUE #166: Implement Decoder–Encoder Architecture for DDPM
### Problem / Goal

We want to implement a decoder–encoder style architecture for DDPM to enable support for arbitrary input formats without requiring separate preprocessing pipelines for different datasets. This will improve flexibility, scalability, and reuse across tasks and datasets, and simplify experimentation and integration with future data sources.

### What does good look like?

- [ ] The model supports switching between different datasets VIA configuration, without touching the codebase
- [ ] New feature does not break existing functionality

### Related context

_No response_

marianovitasari20:
Implement an encoder–decoder style architecture for DDPM to enable support for arbitrary input formats without requiring separate preprocessing pipelines for different datasets. This will improve flexibility, scalability, and reuse across tasks and datasets, and simplify experimentation and integration with future data sources.

With this, the model supports switching between different datasets via configuration, without touching the codebase, via multistage training.

Adds a DDPMProcessor implementing a Denoising Diffusion Probabilistic Model (DDPM) as a processor option in the pipeline, along with the supporting changes needed to train and run it.

Closes #166
W&B link: https://wandb.ai/turing-seaice/train/runs/staf9ajs (this is just an arbitrary run to make sure it runs without further hyperparameter fine-tuning).


Shaerdan:
I will review this one

icenet_mp/models/encode_process_decode.py
```diff
@@ -111,6 +112,14 @@ def __init__(
...
         )
 
+        # DDPM stage trains only the processor (v-prediction in latent space).
+        # The decoder is not on the gradient graph, so freeze it explicitly,
+        # otherwise DDP raises "unused parameters" when running on 2+ GPUs.
+        if isinstance(self.processor, DDPMProcessor):
+            for p in self.decoder.parameters():
+                p.requires_grad = False
+            self.decoder.eval()
+
```
Shaerdan, line 134: if freezing the decoder is a compromise that has to be done to avoid the DDP issue, then maybe it's worth taking a note of this somewhere just in case someone wants to come back and resolve this.

icenet_mp/models/processors/ddpm.py
```diff
@@ -0,0 +1,486 @@
...
+* Autoregressive: generates forecast steps sequentially, updating the history
+  with each predicted target latent.
+* Parallel: generates all forecast steps jointly in a single diffusion
+  process, with forecast steps folded into the channel dimension.
+
+Only the target-latent channels are generated by the diffusion model.
+Non-target channels are not predicted; they are copied from the last history
+frame and carried forward unchanged.
+"""
+
```
Shaerdan, line 19: LGTM, one comment: i notice there is no scale normalisation applied to t0 input; it is fine in general because the deviation of the covariance from the identity would dissipate over ddpm forward pass timesteps and eventually it converges to approximately N(0,1). But since ddpm inherently preserve unit variance -- meaning if you normalise x0 to have unit variance, then all x_i will have exactly unit variance too -- thus if you do the scale normalisation to x0, it could save you the number of timesteps needed to achieve reasonable results. Not binding to this PR but you could give that a try in the future

Shaerdan (approved):
LGTM, I made a few comments for possible changes you can test out in future iterations, not binding to the approval of this PR. Feel free to merge and squash :)

icenet_mp/config/model/ddpm.yaml
```diff
@@ -6,4 +6,4 @@ dropout_rate: 0.05
 mask_type: active
 restrict_range: clamp
 start_out_channels: 64
-timesteps: 150
+timesteps: 1000
```
jemrobinson: Is this deliberate? The 150 change was made so that we could run in a reasonable time on HPC.

icenet_mp/models/processors/ddpm.py
```diff
@@ -0,0 +1,486 @@
...
+        Args:
+            x (TensorNCHW): Combined-latent tensor of shape (B, C_combined, H, W).
+
+        Returns:
+            TensorNCHW: Target-latent slice of shape (B, C_target, H, W).
+
+        """
+        s = self.target_slice_start
+        return x[:, s : s + self.c_target]
```
jemrobinson: This doesn't seem to be used anywhere. Instead there are a couple of instances of `x[:, s : s + self.c_target]`. Could you standardise on one?

icenet_mp/models/processors/ddpm.py
```diff
@@ -0,0 +1,486 @@
...
+            cond = self._flatten_history(cond_window)  # (B, T_hist * C_combined, H, W)
+
+            # Start each forecast step from Gaussian noise.
+            y = torch.randn((b, self.c_target, h, w), device=device)
+
+            # Iteratively denoise from the final diffusion timestep to zero.
+            for t_step in reversed(range(self.timesteps)):
+                t = torch.full((b,), t_step, dtype=torch.long, device=device)
+                pred_v = self.model(y, t, cond)
+                y = self.diffusion.p_sample(y, t, pred_v)
```
jemrobinson: This is duplicated in `_sample_parallel`. Consider factorising this into a separate function (e.g. `_run_reverse_diffusion(self, y, cond) -> torch.Tensor`).

icenet_mp/models/processors/ddpm.py
```diff
@@ -0,0 +1,486 @@
...
+
+        target_ntchw = torch.stack(
+            all_predictions, dim=1
+        )  # (B, T_fcst, C_target, H, W)
+        last_frame = x[:, -1]
+        combined = (
+            last_frame.unsqueeze(1)
+            .expand(b, self.n_forecast_steps, self.c_combined, h, w)
+            .clone()
+        )
```
jemrobinson: This logic is in two other places (`_build_metrics_prediction` and `_sample_parallel`). Again, consider using a helper function here.

icenet_mp/models/encode_process_decode.py
jemrobinson: I think this is a bad idea. This means that there's no way to train the decoder at all, so your output will be dependent on whatever the decoder was initialised to.

Instead, we should require that this model is *only* trained in `multistage` mode, where the decoder is trained separately. I've implemented this change in 0dc8e8d9.

icenet_mp/models/processors/ddpm.py
```diff
@@ -0,0 +1,486 @@
+"""Latent-space DDPM processor.
```
jemrobinson, line 1: Can we add any tests to tests/models/test_processors.py for this processor?

jemrobinson (changes requested):

icenet_mp/models/encode_process_decode.py
marianovitasari20: Thanks both! @jemrobinson, I see your point on the frozen decoder part, thanks for the commit.

icenet_mp/models/processors/ddpm.py
marianovitasari20: Thanks @Shaerdan! I’ve opened a new issue for this for future reference: https://github.com/alan-turing-institute/icenet-mp/issues/438

icenet_mp/config/model/ddpm.yaml
marianovitasari20: Yes, this was deliberate. I changed it back to 1000 so that we can have a proper comparison for DDPM, since the results also degrade noticeably with fewer timesteps. I’m happy to put it back to 150 if 1000 timesteps makes the runs impractical on HPC.

jemrobinson: > the results also degrade noticeably with fewer timesteps.

Interesting. I found that 150 timesteps (which lets us run around 100 epochs on HPC) gives good results but 1000 timesteps (which lets us run around 10 epochs on HPC) gives very bad results.

I'd rather leave this at 150 unless you can show a recent HPC run that performs well with 1000 timesteps.

marianovitasari20: Ahh, that makes sense given the different number of epochs. With the same number of epochs, higher T tends to give better results, but given the HPC runtime constraints, I’ll put it back to 150 in this case. Thanks for flagging this!

icenet_mp/models/encode_process_decode.py
```diff
@@ -140,23 +141,18 @@ def forward(self, inputs: dict[str, TensorNTCHW]) -> TensorNTCHW:
...
-
-        # Add persistence skip connection if requested
-        persistence: TensorNTCHW | None = (
-            inputs[self.output_space.name][
-                :, -1, self.target_variable_indices, :, :
-            ].unsqueeze(1)
-            if self.decoder.skip_connection
-            else None
-        )
+        # Process in latent space: tensor with (batch_size, n_forecast_steps, n_latent_channels_total, latent_height, latent_width)
```
jemrobinson, line 177: Looking at this, I realised that the logic we want here is already in `ProcessorStage`. I've refactored to move that logic from `ProcessorStage` to `EncodeProcessDecode` in bf05d619.

icenet_mp/models/encode_process_decode.py
jemrobinson: I've made an additional simplifying change in 827a5cb

icenet_mp/models/processors/ddpm.py
```diff
@@ -0,0 +1,486 @@
...
+        timesteps: int = 1000,
+        beta_schedule: str = "cosine",
+        kernel_size: int = 3,
+        start_out_channels: int = 64,
+        time_embed_dim: int = 256,
+        dropout_rate: float = 0.1,
+        normalization: str = "groupnorm",
+        activation: str = "SiLU",
+        use_autoregressive: bool = True,
+        target_slice_start: int = 0,
```
jemrobinson: This is too fragile. We should instead calculate where the target channels live (this is deterministic).

icenet_mp/models/encode_process_decode.py
```diff
@@ -110,6 +99,7 @@ def __init__(
         self.processor: BaseProcessor = hydra.utils.instantiate(
```
jemrobinson, line 115: To calculate the target_slice_start on the fly, you want something like this:

```suggestion
        target_slice_start: int | None = None
        offset = 0
        for encoder, input_space in zip(self.encoders, self.input_spaces, strict=True):
            if input_space.name == self.output_space.name:
                target_slice_start = offset
                break
            offset += encoder.data_space_out.channels

        self.processor: BaseProcessor = hydra.utils.instantiate(
```

icenet_mp/models/processors/ddpm.py
```diff
@@ -0,0 +1,486 @@
...
+                f"{c_target} does not fit inside combined latent channels="
+                f"{c_combined}."
+            )
+            raise ValueError(msg)
+
+        self.timesteps = timesteps
+        self.use_autoregressive = use_autoregressive
+
+        # Where the target's channels start in the combined latent. 0 means the target encoder is listed first.
+        self.target_slice_start = target_slice_start
```
jemrobinson: Note that this assumes that the target channels are a continuous slice in latent space. This might not always be true since we might (a) want to predict one channel from one dataset and one from another or (b) apply a convolutional layer (or similar) instead of simply concatenating the channels like we do now. Not a problem right now, but something to consider @marianovitasari20.

icenet_mp/models/processors/ddpm.py
marianovitasari20: Good catch, done in https://github.com/alan-turing-institute/icenet-mp/pull/422/changes/a520a7ccf817eca088c9d09131a0b7263145747b

icenet_mp/models/processors/ddpm.py
marianovitasari20: done in https://github.com/alan-turing-institute/icenet-mp/pull/422/changes/eb0eb60a2f894796dde5492ebb30ec43b3222634

icenet_mp/models/processors/ddpm.py
marianovitasari20: done in https://github.com/alan-turing-institute/icenet-mp/pull/422/changes/1e42b4b58d5530203ae600ca6812d32421477912

icenet_mp/models/processors/ddpm.py
marianovitasari20: Thanks for flagging this @jemrobinson. I’ve opened a new issue here for future reference : https://github.com/alan-turing-institute/icenet-mp/issues/440

icenet_mp/models/processors/ddpm.py
marianovitasari20: Done below! I've added 7 different tests for the DDPM processor.
from https://github.com/alan-turing-institute/icenet-mp/pull/422/changes/436360a403df099d61ca2a673d290489c9046118 to https://github.com/alan-turing-institute/icenet-mp/pull/422/changes/8e357aba01246de89bb1e1ae05927343e1dedf97

jemrobinson:
Confirmed that this runs in multistage with `uv run imp train --multistage model=cnn_ddpm_cnn train.trainer.max_epochs=1` at https://wandb.ai/turing-seaice/train/runs/eylmbpee.

jemrobinson (approved):
Made a couple of simplifying changes. LGTM

PR #428, opened 2026-08-15
Normalize dates before dataset lookup

### LINKED ISSUE #427: Fix error for input plots
### Problem / Goal

When I try running `evaluate` with the option for input plots enabled:

```
defaults:
  - sample
 
evaluate:
  callbacks:
    plotting:
      make_input_plots: true
```
I get the following error:
`IndexError: Date 2024-07-29 not found in the dataset 2024-01-01T12:00:00.000000 to 2024-07-31T12:00:00.000000 every 86400000000 microseconds`

It runs fine when using `make_input_plots: false`

### What does good look like?

- [ ] Check whether this is a issue specific to me
- [ ] Fix the bug

### Related context

_No response_

sylvesterkaczmarek:
Fixes #427.

## Summary

- normalize dates before `SingleDataset.to_index()` lookups
- normalize `get_tchw_slice()` start dates in the same way
- add a regression test covering midnight inputs against datasets whose dates are normalized to noon

## Why

`SingleDataset.dates` normalizes timestamps to noon, but the public lookup methods previously used caller-provided `np.datetime64` values verbatim. Date-only inputs such as `2024-07-29` therefore failed to match an otherwise available `2024-07-29T12:00:00` entry, which breaks evaluation input plotting.

The lookup boundary now applies the same `normalise_date()` convention already used when building the dataset date index.

## Testing

A focused regression test was added. The full repository test suite was not executed locally, so this is opened as a draft for CI validation.


jemrobinson (changes requested):
Thanks for this PR @sylvesterkaczmarek! Could you rebase onto current main and resolve any conflicts? This looks good but isn't mergeable as-is.

PR #431, opened 2026-08-17
Adds some pre-training job feature analysis information

AoifeHughes:
## Summary
Adds an opt-in pipeline for screening input variables before model training. It combines unsupervised diagnostics with spatially resolved predictive importance and produces a consolidated evidence report.
Existing training behaviour and baseline configurations remain unchanged.

## What this adds 
- VIF multicollinearity analysis
- PCA and variable-space EOF diagnostics
- Pairwise correlation analysis
- Spatial feature importance using:
- RandomForestRegressor
- Optional HistGradientBoostingRegressor
- Temporally purged cross-validation
- RF-versus-persistence model qualification
- Grouped permutation importance by physical variable
- Optional add/drop confirmation refits
- Consolidated JSON, CSV and Markdown evidence reports
- Plots for importance, reliability, model quality and spatial-stratum results

## Example usage

```
uv run imp pre-feature-analysis \
  --config-name rf_screening/03_feature_screening \
  platform=<your-platform> \
  data/split=<your-split>
```


docs/src/how-to/vif-analysis.md
```diff
@@ -0,0 +1,111 @@
...
+This loads all variables from the SIC dataset and runs every screening strand, including VIF, against them. Use `++vif.max_samples` to limit the number of timesteps loaded for the VIF strand specifically (useful for large datasets).
+
+### Using multiple datasets (full analysis)
+
+The baseline configs reference three dataset groups: SIC, ERA5 weather, and ARGO float data. To analyse multicollinearity across all input variables:
+
+```bash
+uv run imp pre-feature-analysis --config-name baseline/01_unet.yaml \
+  ++base_path=/Volumes/Storage/ClimateData \
+  'data/datasets=[full_sicnorth_ssmis_25p0km_1979_2024_24h_v2, full_weathernorth_era5_25p0km_1999_2025_24h_v3, full_floatnorth_argo_25p0km_1999_2025_24h_v2]' \
```
jemrobinson, line 38: I think `data=full_south` would do the same thing here.

docs/src/how-to/vif-analysis.md
```diff
@@ -0,0 +1,111 @@
...
+**Important**: You must override `data/datasets` to list all three groups. Passing a single dataset name replaces the entire list rather than adding to it.
+
+### Using specific variables only
+
+To limit analysis to particular variables (faster, less output):
+
+```bash
+uv run imp pre-feature-analysis --config-name baseline/01_unet.yaml \
+  ++base_path=/Volumes/Storage/ClimateData \
+  'data/datasets=[full_sicnorth_ssmis_25p0km_1979_2024_24h_v2]' \
```
jemrobinson, line 51: Ditto `data=full_north`

docs/src/how-to/index.md
```diff
@@ -7,5 +7,8 @@ Step-by-step guides for common tasks.
 - [Train a model](train.md) - run single-stage end-to-end training
 - [Train in stages](train-multistage.md) - pretrain each component separately before finetuning
 - [Evaluate a model](evaluate.md) - evaluate a trained model
+- [Run VIF analysis for multicollinearity](vif-analysis.md) - identify redundant input variables before training
+- [Run feature screening](feature-screening.md) - PCA, EOF, Random Forest importance, correlation, and the consolidated evidence report
```
jemrobinson, line 11: Why are these two different how-tos rather than one? Are they qualitatively different?

docs/src/how-to/index.md
```diff
@@ -7,5 +7,8 @@ Step-by-step guides for common tasks.
 - [Train a model](train.md) - run single-stage end-to-end training
 - [Train in stages](train-multistage.md) - pretrain each component separately before finetuning
 - [Evaluate a model](evaluate.md) - evaluate a trained model
+- [Run VIF analysis for multicollinearity](vif-analysis.md) - identify redundant input variables before training
```
jemrobinson, line 10: Simplify this title so it makes sense to someone who doesn't know what VIF is; e.g. something like "Find input variable correlations"

icenet_mp/config/data/datasets/full_weathernorth_era5_25p0km_1999_2025_24h_v3.yaml
```diff
@@ -0,0 +1,57 @@
+full-weathernorth-era5-25p0km-1999-2025-24h-v3:
```
jemrobinson, line 1: Can we remove this file? We already have a full 1979-2025 if you want to run on the whole dataset and various sample datasets if you want a quick turnaround.

icenet_mp/config/data/split/full_dataset_from_1999.yaml
```diff
@@ -0,0 +1,44 @@
+batch_size: 2
```
jemrobinson, line 1: As above - I'm not sure what this file adds.

icenet_mp/config/data/split/quick_2yr_from_1999.yaml
```diff
@@ -0,0 +1,27 @@
+batch_size: 2
```
jemrobinson, line 1: For quick debug runs, could we use `sample_dataset`?

icenet_mp/config/data/full_north_from_1999.yaml
```diff
@@ -0,0 +1,7 @@
+defaults:
```
jemrobinson, line 1: As above - prefer the existing full and sample datasets

icenet_mp/config/feature_evidence/registry_full_north.yaml
```diff
@@ -0,0 +1,40 @@
+# @package _global_
+feature_evidence:
+  registry:
+    entries:
```
jemrobinson, line 4: Could this be structured as a native YAML list-of-dict e.g.

```yaml
entries:
  - source: sic-ssmis
    variable: ice_conc
    family: sea_ice_concentration
    included: true
  - source: sic-ssmis
    variable: algorithm_standard_uncertainty
    family: sic_uncertainty
    included: true
...
```

icenet_mp/config/rf_screening/01_spatial_rf_screening.yaml
```diff
@@ -0,0 +1,33 @@
+# @package _global_
```
jemrobinson, line 1: We prefer to keep the directory structure matching the way that keys are composed into a config file (e.g. the data/dataset folder containing configs that end up in the data/dataset key) rather than using `@package`.

I think a better solution would be to put the various configurations in a `feature_analysis` folder and then set in them in a local config (with `override /feature_analysis: my_rf_configuration_1`) or at the command line with `feature_analysis=my_rf_configuration_1`)

icenet_mp/data/common_data_module.py
```diff
@@ -90,7 +99,11 @@ def __init__(self, config: DictConfig) -> None:
     def datasets(self) -> dict[str, SingleDataset]:
         """Return a dictionary of dataset group names to SingleDataset objects."""
         return {
-            name: SingleDataset(name, paths)
+            name: SingleDataset(
+                name,
+                paths,
+                variables=sorted(self.dataset_variables.get(name, ())),
```
jemrobinson, line 105: Is this to re-order the variables alphabetically? If so, I imagine this will be slow, as it won't necessarily be how they're stored on disk.

icenet_mp/feature_evidence/registry.py
```diff
@@ -0,0 +1,138 @@
+"""Canonical feature-group registry for the opt-in evidence suite."""
+
+from __future__ import annotations
```
jemrobinson, line 3: Instead of this, just use annotations that are compatible with Python 3.11.

icenet_mp/feature_evidence/registry.py
```diff
@@ -0,0 +1,138 @@
...
+from typing import TYPE_CHECKING
+
+if TYPE_CHECKING:
+    from collections.abc import Mapping, Sequence
+
+    from omegaconf import DictConfig
+
+
+@dataclass(frozen=True)
+class FeatureGroup:
```
jemrobinson, line 15: Why `FeatureGroup` not `Feature` or `InputFeature`?

icenet_mp/input_diagnostics/data.py
```diff
@@ -0,0 +1,287 @@
...
+import logging
+from collections import defaultdict
+from pathlib import Path
+from typing import TYPE_CHECKING
+
+import numpy as np
+from omegaconf import DictConfig, ListConfig
+
+if TYPE_CHECKING:
+    from icenet_mp.data_loaders.single_dataset import SingleDataset
```
jemrobinson, line 22:
```suggestion
    from icenet_mp.data import SingleDataset
```

icenet_mp/input_diagnostics/data.py
```diff
@@ -0,0 +1,287 @@
...
+        group_paths: Mapping of group name to zarr paths (from :func:`resolve_datasets`).
+        group_variables: Mapping of group name to variable lists (from :func:`resolve_datasets`).
+        normalise: Whether to apply the dataset's stored normalisation statistics.
+
+    Returns:
+        Mapping of group name to ``SingleDataset`` instance.
+
+    """
+    # Import here to avoid pulling SingleDataset into module-level imports.
+    from icenet_mp.data_loaders.single_dataset import SingleDataset  # noqa: PLC0415
```
jemrobinson, line 113:
```suggestion
    from icenet_mp.data import SingleDataset  # noqa: PLC0415
```

icenet_mp/input_diagnostics/data.py
```diff
@@ -0,0 +1,287 @@
...
+    Args:
+        group_paths: Mapping of group name to zarr paths (from :func:`resolve_datasets`).
+        group_variables: Mapping of group name to variable lists (from :func:`resolve_datasets`).
+        normalise: Whether to apply the dataset's stored normalisation statistics.
+
+    Returns:
+        Mapping of group name to ``SingleDataset`` instance.
+
+    """
+    # Import here to avoid pulling SingleDataset into module-level imports.
```
jemrobinson, line 112: Is it possible to restructure so that this can be imported at file-level?

icenet_mp/cli/plotting.py
```diff
@@ -0,0 +1,55 @@
+"""Shared CLI helpers for opt-in plotting of spatial RF screening results."""
```
jemrobinson, line 1: Helpers should live in the appropriate module (probably either visualisations or feature_analysis in this case)

icenet_mp/cli/pre_feature_analysis.py
```diff
@@ -0,0 +1,439 @@
+r"""``imp pre-feature-analysis`` — Run all input variable analysis strands together.
```
jemrobinson, line 1: There's too much implementation detail in this file. The cli/<name>.py files should just have the CLI entrypoints plus a few lines creating objects and calling methods on them. The implementation details should be moved to e.g. the feature_analysis module.

icenet_mp/cli/pre_feature_analysis.py
```diff
@@ -0,0 +1,439 @@
...
+        )
+
+        evidence_paths = save_evidence_report(
+            build_evidence_rows(
+                registry,
+                vif_path=output_dir / "vif" / "vif_results.json",
+                pca_path=output_dir / "pca" / "pca_results.json",
+                eof_path=output_dir / "eof" / "eof_results.json",
+                correlation_path=output_dir / "correlations" / "correlations.csv",
+                spatial_rf_path=output_dir / "rf" / "spatial_rf_results.json",
```
jemrobinson, line 410: This is hard-coded to `spatial_rf_results.json` but `rf.mode=scalar` writes to `rf_results.json`.

jemrobinson (changes requested):
I really like the examples you showed us using this code, but I think there's simply too much here for one person to review.

There are around 5500 lines of code, 3000 of tests and 1000 of docs in this PR and the existing repo is around 28k lines, so **this PR increases the total codebase by around 35%!**

Some thoughts on how to make review easier

- Merge `main` into this branch as some module names/paths have changed
- Combine`feature_evidence`, `input_diagnostics` and `input_explainability` into a single module called e.g. `feature_analysis`? You could structure as e.g.
```
icenet_mp/feature_analysis/
    __init__.py
    data.py              # resolve_datasets/build_datasets 
    diagnostics/          # was input_diagnostics/ (vif, pca, eof, correlation)
    explainability/        # was input_explainability/ (base, rf, spatial_rf)
    evidence/              # was feature_evidence/ (registry, report)
    runner.py              # _run_all_strands + strand dispatch, moved out of cli/pre_feature_analysis.py
```
- a shared `write_report(result, output_dir)` helper would remove duplication across `vif.py`, `pca.py`, `eof.py`, `rf.py`
- a shared figure saver (ideally integrated into the visualisations module and used by the figure-writing code there) would remove duplication in `plots.py`, `explainability/base.py`, `diagnostics/correlation.py`.
- `rf.py` and `spatial_rf.py` are both ~1000 lines! Factorising out report formatting would help focus these files on the modelling logic.

Ways to take this forward

- consider splitting this into several sequential PRs (waiting to merge each one before preparing the next one):
  - **data loading:** `icenet_mp/feature_analysis/data.py` with resolve_datasets/build_datasets. Can discuss here whether it's possible to reduce duplication with `CommonDataModule`
  - **diagnostics:** `icenet_mp/feature_analysis/diagnostics/` allows for a review of just the statistical implementation of `vif.py`, `pca.py`, `eof.py`, `correlation.py`
  - **explainability:** `icenet_mp/feature_analysis/explainability/` base + scalar RF
  - **spatial RF:** `spatial_rf.py`, `spatial_rf_plots.py,`, `visualisation/plotting.py` (I think this is the most complex piece)
  - **Feature-evidence:** registry + report 
  - **Wrap-up**: everything under `cli/` + docs and demo.
- alternatively: prepare an MVP PR that can be reviewed in one pass
- alternatively: convert the statistical implementation and report preparation into a standalone package that we can include as a dependency (and then we'll just be calling it with the appropriate inputs/outputs/CLI interface).

sylvesterkaczmarek (commented):
Also, I still think the train/validation boundary should be consistent across all screening strands. RF is restricted to data.split.train, while VIF/PCA/EOF/correlation can use the full available dataset. Since these feed the same evidence report, that can leak held-out-period distribution information into feature-selection decisions. I would restrict all diagnostic matrices to the training ranges and record those ranges in the provenance.

AoifeHughes:
Closing as being supersceeded by the smaller: https://github.com/alan-turing-institute/icenet-mp/tree/ah/rf-feature-importance-v2 branch

PR #439, opened 2026-08-18
Normalize dates before dataset lookup

### LINKED ISSUE #427: Fix error for input plots

sylvesterkaczmarek:
Fixes #427.

## Summary

- Normalize dates before `SingleDataset.to_index()` lookups.
- Apply the same normalization to `get_tchw_slice()` start dates.
- Add a regression test covering date-only inputs against dataset dates normalized to noon.

## Why

`SingleDataset.dates` normalizes timestamps to noon, while the lookup methods previously used caller-provided `np.datetime64` values directly. A date-only value such as `2020-01-01` could therefore fail to match the corresponding normalized `2020-01-01T12:00:00` dataset entry.

## Testing

Added a focused regression test covering both `to_index()` and `get_tchw_slice()`.

This branch has also been rebased onto current `main` following the review on #428.

### DIFF
--- a/tests/data/test_date_lookup_normalization.py
+++ b/tests/data/test_date_lookup_normalization.py
@@ -8,6 +8,7 @@ from icenet_mp.data.single_dataset import SingleDataset
 def test_date_lookups_normalise_to_noon(
     mock_dataset_non_normalized_times: Path,
 ) -> None:
+    """Confirm that a midnight datetime is normalised during lookup."""
     dataset = SingleDataset(
         name="test_normalized",
         input_files=[mock_dataset_non_normalized_times],
--- a/tests/data/test_date_lookup_normalization.py
+++ b/tests/data/test_date_lookup_normalization.py
@@ -1,0 +1,21 @@
+from pathlib import Path
+
+import numpy as np
+
+from icenet_mp.data.single_dataset import SingleDataset
+
+
+def test_date_lookups_normalise_to_noon(
+    mock_dataset_non_normalized_times: Path,
+) -> None:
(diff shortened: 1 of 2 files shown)


jemrobinson (approved):
LGTM

tests/data/test_date_lookup_normalization.py
```diff
@@ -0,0 +1,21 @@
+from pathlib import Path
+
+import numpy as np
+
+from icenet_mp.data.single_dataset import SingleDataset
+
+
+def test_date_lookups_normalise_to_noon(
+    mock_dataset_non_normalized_times: Path,
```
jemrobinson, line 9: Add a docstring here to make the linter happy :)

tests/data/test_date_lookup_normalization.py
```diff
@@ -0,0 +1,21 @@
...
+) -> None:
```
jemrobinson, line 10:
```suggestion
) -> None:
    """Confirm that a midnight datetime is normalised during lookup."""
```

Issue #346, opened 2026-07-02
Check ConvNormAct block stacking


jemrobinson:
### Problem / Goal

@Shaerdan raised a question about whether stacked `ConvNormAct` blocks in the encoder/decoder might result in vanishing gradients without skip connections. We should investigate this.

### What does good look like?

- [ ] Understand whether current `ConvNormAct` encoder/decoder pairs have this failure mode

### Related context

Raised on #333.

PR #441, opened 2026-08-19
Add gradient connectivity smoke test for stacked ConvNormAct blocks

sylvesterkaczmarek:
Smoke test for #346.

Exercise the current PiecewiseEncoder/PiecewiseDecoder path with three ConvNormAct subblocks on both sides. Backpropagate through the full pair and verify that the input and every convolution receive finite, non-zero gradients.

This is intentionally a gradient-connectivity smoke test. It does not claim to measure or rule out vanishing-gradient behaviour in deeper or differently normalised networks.

### DIFF
--- a/tests/data/test_single_dataset.py
+++ b/tests/data/test_single_dataset.py
@@ -9,13 +9,6 @@ from icenet_mp.data.single_dataset import SingleDataset
 from icenet_mp.types import DataSpace
 
 
-class MockAnemoiDataset:
-    def __init__(self, channels: int, height: int, width: int) -> None:
-        """A mock Anemoi dataset for testing purposes."""
-        self.shape = (1, channels, height * width)
-        self.field_shape = (height, width)
-
-
 class TestSingleDataset:
     def test_name(self) -> None:
         dataset = SingleDataset(name="test_dataset", input_files=[])
--- a/tests/models/test_piecewise_encode_decode.py
+++ b/tests/models/test_piecewise_encode_decode.py
@@ -71,3 +71,54 @@ class TestPiecewiseEncodeDecode:
         )
         output_ntchw = decoder.rollout(latent_ntchw, None)
         assert torch.equal(input_ntchw, output_ntchw)
+
+    def test_stacked_conv_blocks_propagate_gradients(self) -> None:
+        """Verify the current stacked encoder/decoder conv blocks receive gradients."""
+        torch.manual_seed(0)
+        n_history_steps = 2
+        input_ntchw = torch.randn(2, n_history_steps, 2, 16, 16, requires_grad=True)
+        input_space = DataSpace(name="input", channels=2, shape=(16, 16))
+
+        encoder = PiecewiseEncoder(
+            conv_activation="SiLU",
+            conv_subblocks_initial=3,
+            conv_subblocks_final=3,
+            data_space_in=input_space,
+            latent_space=(4, 4),
+        )
+        decoder = PiecewiseDecoder(
+            conv_activation="SiLU",
+            conv_subblocks_initial=3,
+            conv_subblocks_final=3,
+            data_space_in=encoder.data_space_out,
+            data_space_out=input_space,
+            restrict_range="none",
+            use_final_normalisation=False,
+            use_hann_window=False,
+        )
+
+        latent_ntchw = encoder.rollout(input_ntchw)
+        output_ntchw = decoder.rollout(latent_ntchw, None)
+        output_ntchw.square().mean().backward()
+
+        assert input_ntchw.grad is not None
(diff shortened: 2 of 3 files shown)


Shaerdan (changes requested):
This would be useful as a smoke test, but it cannot be used to prove that ConvNormAct does not have the well-known vanishing gradient issue in deep CNNs, because that issue is very much depth- and normalisation-dependent.

From what I see here, this PR is one forward/backward at random initialisation on a toy using an encoder-decoder pair with nothing between them, i worry it might not be comparable enough to the architecture we use in the real runs. That toy setup might well be a lucky setup for not having gradient issues, so it is probably better considered a smoke test, that it shows there is no gradient graph disconnection, rather than testing the potential architectural tendency of the gradient to become more faint going deeper (despite the gradient graph being correct).

A bigger issue is that this test compares raw weight gradient norms, which gives the wrong stats comparison: it compares raw weight-gradient norms across layers with 36 vs 236,196 elements. Concretely, with the parameters this test uses, the encoder holds three `Conv2d(2, 2, 3, 3)` (36 elements each) and three `Conv2d(162, 162, 3, 3)` (236,196 elements each). Running the test and printing both statistics:

| layer | elements | raw norm | RMS per element |
| --- | --- | --- | --- |
| `Conv2d(2, 2, 3, 3)` | 36 | 3.97e-1 | 6.61e-2 |
| `Conv2d(162, 162, 3, 3)` | 236,196 | 2.75e+0 | 5.66e-3 |

With the raw norm, the wide convs look ~7x better, but by per-element magnitude they are actually ~12x weaker. The comparison flips. The arithmetic is: a norm = sqrt(N) * RMS, so the element count alone contributes a factor of sqrt(236196/36) = 81 to the norm ratio. The observed ratio is only 6.9, which means the per-element magnitude is 81/6.9 ~ 12x smaller. The stats is therefore dominated by the number of entries rather than by the gradient magnitude.

A second reason weight gradients are the wrong object here: `grad_W L = delta * a^T` mingles the backprop signal `delta` with the forward activation `a`, so a layer with large activations and a dying backprop signal can still show a healthy weight-gradient norm. Vanishing gradient is a about `delta`, so a plot of `||dL/da||` per element as a function of depth would have been better.

And lastly, but probably most importantly, the threshold set in the test cannot fail for the reason the test is named after. `torch.finfo(torch.float32).eps` is 1.19e-7, so `norms.min() > eps * norms.max()` only fails when the spread of gradient norms across layers exceeds 8.4e6. On these modules the measured spread is roughly 5-12x (min/max = 8.4e-2 encoder, 1.1e-1 decoder at 3+3 sub-blocks; 1.3e-1 / 1.8e-1 at the production 1+1 setting), so the assertion passes with roughly a 1e6 margin. It is therefore an underflow/denormal detector rather than a vanishing-gradient detector.

To demonstrate this: a bare ConvNormAct stack with `norm_type="none"` and PyTorch's default `Conv2d` init produces the exact failure mode this test is meant to detect and warn against.

| depth | min/max ratio | trainable? | this assertion |
| --- | --- | --- | --- |
| 8 | 1.3e-4 | marginal | PASSES |
| 10 | 2.1e-5 | no | PASSES |
| 12 | 1.8e-7 | no | PASSES |
| 14 | 1.4e-8 | no | FAILS |

In this case the gradients decay by 7 orders of magnitude over its depth (untrainable), but the threshold used in this test passes it as healthy.

I therefore recommend renaming the test to reflect that it is a smoke test for gradient graph connectivity, rather than a test of the vanishing gradient suspicion.

On the underlying suspicion raised in the original #346, I do think BatchNorm prevents it eventually. But people still use a bridge/skip connection from the first ConvNormAct to the last within the same level of a UNet encoder or decoder arm, not just between encoder and decoder. One of the tests I did showed that when `norm_type` is set to "none" the vanishing gradient does indeed happen. So I think it may be best to add the bridge with a residual path that is on by default when `norm_type="none"', at least this could work as a safety fallback.

One implementation note if we do that: CommonConvBlock first sub-block changes the channel count, and several call sites rely on that (PiecewiseDecoder's initial block maps data_space_in.channels to input_channels_required, and UNetProcessor maps `channels[i] -> channels[i+1]`). So a simple `x + F(x)` probably will not typecheck there, meaning the bridge needs either a 1x1 projection shortcut for the channel-changing case, or to span only sub-blocks 2..n, where the channel count is already constant.

Happy to make this a separate PR together with a validation test designed to measure the depth gradient trend.

Shaerdan (approved):
LGTM, the bridge/skip connection between  ConvNormAct will be discussed as a separate issue and the gradient safety check remains open, this PR is renamed as a smoke test for gradient graph connectivity. Approved

PR #442, opened 2026-08-19
Restore standalone dataset input plotting

### LINKED ISSUE #92: Add some visualisations when running dataset inspect
### Problem / Goal

We should add some basic views of the input data at the point we run dataset inspect.

### What does good look like?

- [ ] Dataset visualisations include in `dataset inspect` command

### Related context

_No response_

sylvesterkaczmarek:
Closes #92.

Add `imp datasets plot` to save one static PNG per variable for a selected timestep. `--dataset` limits output to one configured dataset, while omitting it plots all configured datasets. Reuse the current SingleDataset and plotting infrastructure, add focused output/range/CLI tests, and document the command.


jemrobinson (approved):
Thanks @sylvesterkaczmarek! I've moved the logic into the `visualisations` module as we prefer the `cli` module not to contain implementation details. I also added the ability to make video as well as static plots.

PR #449, opened 2026-08-21
Add uncertainty-aware SIC visualisation

### LINKED ISSUE #404: Add a visualisation that incorporates data uncertainty
### Problem / Goal

We have a `total_standard_uncertainty` variable in SSMIS. We should consider including this in our visualisations of ground-truth vs. prediction (e.g. showing the different in terms of `z`).

### What does good look like?

- [ ] Add an uncertainty visualisation

### Related context

_No response_

sylvesterkaczmarek:
## Summary

Adds an uncertainty-aware sea ice concentration visualisation using the SSMIS `total_standard_uncertainty` field.

## Changes

- compute standardised error as `(ground_truth - prediction) / total_standard_uncertainty`
- add a static z-score map with a symmetric diverging scale
- log the uncertainty view alongside the existing static prediction plots when SSMIS uncertainty is available
- preserve existing plotting behaviour when the uncertainty field is absent
- mask undefined values where uncertainty is zero or non-finite
- add focused tests for the standardised-error calculation and rendering path

Closes #404.


icenet_mp/callbacks/plotting_callback.py
```diff
@@ -165,12 +222,14 @@ def make_plots(
         channel_names = getattr(pl_module, "channel_names", ["sea-ice-concentration"])
 
         if self.make_static_plots:
+            uncertainties = self.load_target_uncertainties(dataset, dates)
```
jemrobinson, line 239: Wrap this in a try/except to ensure that failure to load the uncertainties doesn't crash a training job.

jemrobinson (approved):
LGTM. Did some light refactoring to move functions to the locations appropriate for the codebase.

Issue #62, opened 2025-08-18
Add more tests


jemrobinson:
### Problem / Goal

Our current test suite doesn't cover much of the codebase, we should increase this.

### What does good look like?

- [ ] Tests for callbacks
- [ ] Tests for cli
- [ ] Tests for config
- [ ] Tests for data_loaders
- [ ] Tests for data_processors
- [ ] Tests for evaluation
- [ ] Tests for training
- [x] Tests for visualisations
- [ ] Tests for types
- [ ] Tests for utils

PR #451, opened 2026-08-21
Add utility module test coverage

sylvesterkaczmarek:
## Summary

Adds focused unit-test coverage for `icenet_mp.utils` as a small contribution toward #62.

## Coverage added

- NumPy datetime to UTC datetime conversion
- mask-directory path construction
- accelerator/device naming and CUDA fallback behaviour
- timestamp formatting
- W&B run lookup fallback
- date normalisation to noon
- string/list normalisation with `to_list`

This PR is test-only and does not change production behaviour.

Part of #62.


jemrobinson (approved):
LGTM - tidied some codestyle and added some extra tests.

Issue #366, opened 2026-07-09
Handle missing dates


jemrobinson:
### Problem / Goal

At the moment our training data is constrained as the union of those dates for which *all* datasets have valid data. This means that include Argo float data substantially reduces the amount of training data

### What does good look like?

- [ ] Implement a better way of handling missing dates (e.g. provide a dummy tensor with -1 everywhere)
- [ ] Confirm that this behaves sensibly during training.

### Related context

- Related to #314

PR #452, opened 2026-08-21
Handle missing conditioning dates without dropping samples

sylvesterkaczmarek:
## Summary

Adds opt-in handling for missing conditioning dates so sparse inputs such as Argo no longer force otherwise valid target windows to be dropped.

## Changes

- add `SingleDataset.get_tchw_with_fill()` for normalized reads with a `-1` missing-data sentinel
- add `CombinedDataset.allow_missing_inputs` while preserving strict date intersection by default
- keep target history and forecast ground truth mandatory
- wire the option through `CommonDataModule`
- enable it for the sample and full north/south configurations that include sparse Argo observations
- add regression tests for missing conditioning data and target-window validation
- document the configuration behavior

The dataloader behavior is covered by tests. A full real-data training run is still needed to complete the training-validation part of the issue.

Part of #366.


jemrobinson (changes requested):
What impact does this have on performance? It would be good to see both (a) impact on training/inference time and (b) impact on prediction quality. Are you interested in looking into this @sylvesterkaczmarek ?

Issue #254, opened 2026-04-20
Streamline data loading


jemrobinson:
### Problem / Goal

During data loading, we load individual files multiple times.

### What does good look like?

- [ ] Streamline data loading so that we minimise the number of times that files are loaded

jemrobinson:
Done in https://github.com/alan-turing-institute/icenet-mp/pull/263

PR #454, opened 2026-08-21
Add regression coverage for cached dataset loading

sylvesterkaczmarek:
## Summary

Adds regression coverage for the class-level Anemoi dataset cache used by `SingleDataset`.

## Coverage added

- repeated loads of the same fileset call `open_dataset` only once
- `SingleDataset` subsets reuse the same cached source dataset instead of reopening the files

This protects the existing load-minimization behavior from regressions.

Part of #254.


jemrobinson (approved):
LGTM

PR #459, opened 2026-08-21
Add ActivationSaver callback test coverage

sylvesterkaczmarek:
## Summary

Adds focused unit coverage for the `ActivationSaver` evaluation callback.

## Coverage added

- invalid layer paths fail with an actionable error
- repeated layer calls keep only the first activation for a batch
- tensor inputs and captured activations are persisted in the expected batch payload
- metadata is written at test end
- hooks are detached after evaluation
- an empty layer list remains a no-op

This PR is test-only and does not change runtime behavior.

Part of #62.


jemrobinson (approved):
LGTM after refactoring.

Issue #150, opened 2025-11-28
Improve evaluation process


IFenton:
### Problem / Goal

Improve the evaluation process, particularly in terms of the metrics calculated and the W&B plots produced

### What does good look like?

- [ ] All subissues completed

PR #463, opened 2026-08-21
Clarify evaluation metrics and example videos

sylvesterkaczmarek:
## Summary

Clarifies how per-forecast-day metrics and example videos should be interpreted during evaluation.

## Changes

- explain that forecast-day metrics are aggregated across evaluation start dates
- explain that videos use a small set of representative start dates
- distinguish lead-time performance charts from example forecast sequences
- make no runtime or evaluation-code changes

Part of #150.


jemrobinson (approved):
LGTM

Issue #206, opened 2026-01-29
Decide which sea ice concentration datasets to use


jemrobinson:
## Currently using
- OSISAF SSMIS 25km grid **1978-2025** (https://osi-saf.eumetsat.int/products/osi-450-a1 and https://osi-saf.eumetsat.int/products/osi-430-a)

## Available
- OSISAF SSMIS 10km grid **2005-now** (https://osi-saf.eumetsat.int/products/osi-401-d)
- OSISAF AMSR2 10km grid **2016-now** (https://osi-saf.eumetsat.int/products/osi-408-a)
- OSISAF AMSR2 25km grid **2025-now** (https://osi-saf.eumetsat.int/products/osi-438)
- OSISAF AMSR2 25km grid **2002-2025** (https://osi-saf.eumetsat.int/products/osi-458)

### What does good look like?

- [ ] Decide whether we want one or more of these datasets

jemrobinson:
As of #397 we decided to use 25 km grids for our main dataset and consider 10 km grids as a future extension.

PR #465, opened 2026-08-21
Document the current sea-ice dataset selection

sylvesterkaczmarek:
## Summary

Documents the 25 km OSI SAF sea-ice concentration record already represented by the current full dataset configuration and clarifies how alternative higher-resolution products should be evaluated.

## Changes

- document the OSI-450-a1, OSI-458 and OSI-438 transitions used by the current record
- explain why the long 25 km record remains a practical baseline
- distinguish 10 km products as alternatives requiring an explicit resolution/training-coverage comparison
- point to the north and south full dataset descriptors
- add the guide to the documentation navigation

Documentation only; no data ingestion or runtime behaviour changes.

Part of #206.


jemrobinson (approved):
LGTM. Expanded to cover ERA5 and Argo float.

Issue #228, opened 2026-02-23
Simplify plotting module


jemrobinson:
### Problem / Goal

The plotting module has multiple indirections, is hard to change and complicated to test. We should simplify it.

### What does good look like?

- [ ] Simpler plotting module

PR #466, opened 2026-08-21
Add Plotter orchestration contract tests

sylvesterkaczmarek:
## Summary

Adds focused characterization tests around `Plotter` orchestration so the plotting module can be simplified without accidentally changing its logging contract.

## Coverage added

- prefixed and default static logging namespaces
- explicit and fallback output channel names
- video-buffer rewind before logger handoff
- configured video-format forwarding
- hemisphere propagation into `PlotSpec`

This PR is test-only and does not change plotting or runtime behaviour. It provides a small safety net for the refactoring requested in #228.

Part of #228.


jemrobinson (approved):
LGTM. Refactored to use Protocols.

PR #471, opened 2026-08-21
Add train CLI regression coverage

sylvesterkaczmarek:
## Summary

Adds focused regression coverage for the `imp train` command.

## Coverage added

- default training composes the requested Hydra config and passes it to `ModelService`
- single-stage training remains the default
- `--multistage` is forwarded correctly
- `--checkpoint-dir` is resolved before being passed to the training service

This PR is test-only and changes no production behaviour.

Part of #62.


jemrobinson (approved):
LGTM

Issue #438, opened 2026-08-18
Explore scale-normalizing x0 for DDPM


marianovitasari20:
### Problem / Goal

Following up on @Shaerdan's review comment on #422: "i notice there is no scale normalisation applied to t0 input; it is fine in general because the deviation of the covariance from the identity would dissipate over ddpm forward pass timesteps and eventually it converges to approximately N(0,1). But since ddpm inherently preserve unit variance -- meaning if you normalise x0 to have unit variance, then all x_i will have exactly unit variance too -- thus if you do the scale normalisation to x0, it could save you the number of timesteps needed to achieve reasonable results. Not binding to this PR but you could give that a try in the future"

Not blocking any current work, just worth experimenting with in a future iteration.

### What does good look like?

- [ ] A short experiment comparing sample quality/timesteps needed with vs. without normalizing `x0` to unit variance.

PR #478, opened 2026-08-24
Add configurable x0 scaling for DDPM experiments

sylvesterkaczmarek:
## Summary

- add `x0_scale` directly to `DDPMProcessor`, defaulting to `1.0`
- normalize clean target latents before forward diffusion
- restore predictions to the original latent scale before metrics/decoder use
- reject non-positive scale values
- keep existing DDPM behaviour unchanged when the parameter is omitted

## Testing

- `uv run pytest -q -n 0 tests/models/test_ddpm_scaling.py` - 4 passed
- `uvx ruff check` on changed Python files
- `uvx ruff format --check` on changed Python files

Part of #438.


jemrobinson (changes requested):
I think this would be better as a parameter to DDPMProcessor (with a default of 1) rather than a separate processor.

Issue #129, opened 2025-10-17
Add climatology ice edge contour to evaluation plots


aranas:
### Problem / Goal

Based on our pre-computed climatology baseline we can evaluate our model performance by comparing:
- predicted ice edge against climatology ice edge
- SIC prediction error against climatology error

### What does good look like?

- [ ] in ice edge contour plot options to add climatology ice edge as a differently coloured contour trace
- [ ] in accuracy/error line plots (eg error profile over lead times or different init months) option to add a climatology based error
- [ ] in heatmap plots generate a separate heatmap for climatology baseline to compare performance across different init months and lead times

### Related context

- related issue for climatology baseline generation https://github.com/alan-turing-institute/ice-station-zebra-project-board/issues/83
- see also https://github.com/alan-turing-institute/ice-station-zebra-project-board/issues/82

jemrobinson:
Add this as e.g. a green contour to the truth plot in the truth-prediction-difference figure.

PR #480, opened 2026-08-24
Add climatology ice-edge overlay to static prediction plots

sylvesterkaczmarek:
## Summary

- allow `plot_static_prediction` to accept an optional climatology SIC array
- when supplied, overlay only the climatology 15% ice edge on the prediction panel as a distinct reference contour
- leave existing plots unchanged when no climatology is supplied
- validate climatology/prediction shape compatibility
- add focused unit and integration coverage for the overlay

This deliberately does not duplicate the prediction/ground-truth ice-edge work being developed in #396. It adds the climatology reference requested by #129.

## Side-by-side

Existing output versus opt-in climatology edge

The left side is the existing output. The right side is the same plot with the optional climatology contour supplied.

## Testing

- regression coverage for the 15% climatology contour and distinct colour
- regression coverage for mismatched climatology shapes
- integration test confirming static prediction rendering with the optional overlay

Part of #129.


jemrobinson (commented):
Is this useful on its own or does it make the comparison plots noisier? Can you show a side-by-side comparison?

PR #481, opened 2026-08-24
Add direct tests for utility helpers

sylvesterkaczmarek:
## Summary

Add direct coverage for `icenet_mp.utils`, including:

- NumPy datetime to UTC datetime conversion
- preprocessing mask paths
- CPU, MPS, CUDA and XPU device-name handling, including unavailable-device fallbacks
- UTC timestamp formatting
- WandB run discovery
- date normalisation
- string/list conversion

## Testing

- adds focused unit tests only; production behaviour is unchanged

Part of #62.


jemrobinson (approved):
LGTM

PR #483, opened 2026-08-24
Add EMA and unconditional checkpoint callback tests

sylvesterkaczmarek:
## Summary

Adds focused test coverage for callback paths not covered by the existing ActivationSaver test PR:

- EMA update scheduling by step and epoch
- parameterless-model handling
- delegation to Lightning's weight-averaging hooks for parameterised modules
- unconditional checkpoint directory handling and path construction
- distributed filepath broadcast and checkpoint save calls
- opt-in checkpoint saving at training end

## Testing

- test-only change; production behaviour is unchanged

Part of #62.

### DIFF
--- a/tests/callbacks/conftest.py
+++ b/tests/callbacks/conftest.py
@@ -1,16 +1,32 @@
 from unittest.mock import MagicMock
 
 import pytest
+import torch
 from lightning import LightningModule, Trainer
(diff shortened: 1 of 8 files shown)


jemrobinson (approved):
LGTM after refactor

Issue #40, opened 2025-07-30
Dealing with non gridded data


aranas:
This issue concerns with how we will handle non gridded data into the pipeline.

Possible solution is the popular [setconv](https://aclanthology.org/2020.emnlp-main.98/) to transform non gridded data to gridded data.

Consideration for non-gridded data directly: 
Use of transformers treat irregularly spaced data as a sequence of tokens, and use self-attention mechanisms to process them without relying on a grid-based structure (Wang et al., 2025 _Local Off-Grid Weather Forecasting with Multi-Modal Earth Observation Data_). The paper claims transformer is one of the best performing models for off-grid local weather forecast due to its flexible attention mechanism.

TBC

aranas:
Alternative approach suggested by @louisavz 

Consideration for non-gridded data directly: 
Use of transformers treat irregularly spaced data as a sequence of tokens, and use self-attention mechanisms to process them without relying on a grid-based structure (Wang et al., 2025 _Local Off-Grid Weather Forecasting with Multi-Modal Earth Observation Data_). The paper claims transformer is one of the best performing models for off-grid local weather forecast due to its flexible attention mechanism.

PR #484, opened 2026-08-24
Add fixed-coordinate SetConv encoder for sparse observations

sylvesterkaczmarek:
## Summary

- add a `SetConvEncoder` that maps sparse or irregular observations onto a regular latent grid
- use spherical Gaussian kernel weighting through the existing latitude/longitude provider interface
- include observation-density features and a configurable, optionally learnable kernel length scale
- ignore non-finite observations when aggregating sparse inputs
- export the encoder and add focused unit tests

## Scope

This first implementation assumes sensor coordinates are fixed across timesteps, matching the current pipeline coordinate interface and the limitation discussed in #40. Supporting moving sensor coordinates remains future work.

## Testing

- added rollout shape and gradient coverage
- added missing-observation coverage
- added interpolation coverage for matching source/target coordinates
- added coordinate validation and invalid-length-scale coverage

Part of #40.

### DIFF
--- a/icenet_mp/geotools/__init__.py
+++ b/icenet_mp/geotools/__init__.py
@@ -7,7 +7,7 @@ from .grid_factory import (
     epsg_6931_builder,
     epsg_6932_builder,
 )
-from .reproject import nearest_neighbour_indices
+from .reproject import nearest_neighbour_indices, pairwise_haversine_distances
 
 grid_factory = GridFactory()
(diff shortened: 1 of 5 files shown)


jemrobinson (changes requested):
1. We already have tools to convert lat-long pairs into distances in geotools, without going lat->deg->pos.
2. This encoder isn't useful on its own as we don't currently have a fixed-coordinate dataset to run over. Can you add a minimal example (perhaps using an Anemoi trajectory dataset)?

Issue #406, opened 2026-07-31
Investigate different schedulers


jemrobinson:
### Problem / Goal

Currently we are using `CosineAnnealingLR` as our default scheduler which has an annoying feature that LR starts to increase again if you go past `T_max`, so you need to be careful that `T_max` is aligned with your expected number of epochs.

We should look at other schedulers (e.g. a warm-up scheduler, that ramps up then back down again).

### What does good look like?

- [ ] See which schedulers work best for which models.

sylvesterkaczmarek:
I ran a paired comparison rather than keeping this as scheduler-only scaffolding.

Setup: current `synthetic` config / `quick-test` UNet, seed 123, 30 epochs, same AdamW settings and data, Apple MPS, plotting disabled. Compared the existing `CosineAnnealingLR` against this 5-epoch linear-warmup + cosine schedule.

Results:
- existing cosine: best validation loss **0.02486** at epoch 29 (final 0.02486)
- warmup cosine: best validation loss **0.02846** at epoch 13 (final 0.03223)

The warmup run converged faster early, but the existing scheduler overtook it and finished about 12.7% lower on best validation loss. This is only one model/seed, so it is not a global scheduler recommendation, but it does mean I do not have evidence that adding this scheduler improves the current quick-test setup. I would rather close this implementation than add an unproven option. A broader #406 study can compare schedulers across representative models/seeds before introducing another default-supported scheduler.

PR #487, opened 2026-08-24
Add non-restarting warmup cosine scheduler

sylvesterkaczmarek:
## Summary

Adds an opt-in warm-up cosine learning-rate scheduler for the scheduler investigation in #406.

The scheduler:
- linearly warms the learning rate from a configurable start factor to the base learning rate
- applies cosine decay after warm-up
- clamps at `eta_min` after the configured training horizon instead of increasing again
- is available as `train/scheduler=warmup_cosine`
- leaves the existing default scheduler unchanged

## Testing

Adds focused tests for:
- linear warm-up behaviour
- cosine decay
- staying at `eta_min` after the training horizon without a cosine restart
- invalid scheduler configuration

Part of #406.

### DIFF
--- a/icenet_mp/config/train/scheduler/warmup_cosine.yaml
+++ b/icenet_mp/config/train/scheduler/warmup_cosine.yaml
@@ -1,0 +1,6 @@
+# Linear warm-up followed by cosine decay that stays at eta_min after training.
+_target_: icenet_mp.schedulers.WarmupCosineAnnealingLR
+warmup_epochs: 5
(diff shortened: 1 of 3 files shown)


jemrobinson (changes requested):
The aim of #406 is to define which different schedulers to consider and to come up with a data-based recommendation. Although adding additional schedulers might help with that, it's not useful on its own. Are you interested in running some comparisons between different schedulers @sylvesterkaczmarek ?

Issue #155, opened 2025-12-03
Allow easy integration of pretrained embeddings


aranas:
### Problem / Goal

Currently we learn directly from satellite data products. There might be value in learning from high-dim environmental embeddings such as presto, [galileo](https://arxiv.org/abs/2502.09356), alphaearth

### What does good look like?

- Adapt pipeline to be able to ingress embeddings (of various formats?)
- Explore benefits of training on embeddings vs L4 data products
- what good looks like needs to be further explored

PR #488, opened 2026-08-24
Add pretrained embedding encoder

sylvesterkaczmarek:
## Summary

- add `PretrainedEmbeddingEncoder` for externally generated gridded environmental embeddings
- preserve embeddings exactly when spatial shape and channel count already match the shared latent space
- optionally resize embedding maps without adding learned spatial convolutions
- optionally project embedding channels with a 1x1 convolution
- support optional normalisation through the existing normalisation factory
- export the encoder and add focused passthrough, resize, projection, rollout and validation tests

## Motivation

This provides a minimal pipeline integration point for the pretrained environmental embeddings discussed in #155 without forcing them through a spatial feature extractor that may unnecessarily relearn an upstream representation.

## Testing

- exact passthrough coverage
- spatial resize coverage
- channel projection and gradient coverage
- rollout shape coverage
- invalid interpolation-mode validation

Part of #155.

### DIFF
--- a/icenet_mp/models/encoders/__init__.py
+++ b/icenet_mp/models/encoders/__init__.py
@@ -3,6 +3,7 @@ from .cnn_encoder import CNNEncoder
 from .deep_compression_encoder import DeepCompressionEncoder
 from .naive_linear_encoder import NaiveLinearEncoder
 from .piecewise_encoder import PiecewiseEncoder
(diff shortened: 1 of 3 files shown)


jemrobinson (changes requested):
The aim of #155 isn't just to add the scaffolding for including pre-trained embeddings but to do this in practice and demonstrate whether it helps with our predictions. Are you interested in looking into this @sylvesterkaczmarek ?

PR #493, opened 2026-08-24
Add ModelService multistage checkpoint tests

sylvesterkaczmarek:
## Summary

Adds focused regression coverage for `ModelService` multistage configuration and checkpoint helpers.

The tests cover:
- rejecting a multistage checkpoint directory during standard single-stage training
- merging stage-specific training overrides while retaining common settings
- saving a deterministic stage checkpoint when no callback has selected a best checkpoint
- moving a single callback-selected best checkpoint to the stage-specific path
- synchronising ranks after moving the selected checkpoint
- rejecting ambiguous checkpoint selection when multiple best paths are present

This is a test-only change; production behaviour is unchanged.

## Testing

Part of #62.


jemrobinson (approved):
Added some additional missing tests.

PR #500, opened 2026-08-24
Add LocalFileLogger regression coverage

sylvesterkaczmarek:
## Summary

Adds focused regression coverage for `LocalFileLogger`, which currently has no dedicated tests.

Coverage includes:

- logger name, version and save-directory metadata
- JSONL metric appends and numeric coercion
- unique image filenames across repeated logging calls
- sanitisation of media keys for filesystem-safe filenames
- explicit image-step handling
- warning and skip behaviour for non-image objects
- video-buffer rewind and requested format handling
- default MP4 output and unique video call indices

This is a test-only change; production behaviour is unchanged.

## Testing

Part of #62.


jemrobinson (approved):
LGTM

PR #509, opened 2026-08-25
Export evaluation predictions to NetCDF

### LINKED ISSUE #507: Add functionality to extract data from predictions
### Problem / Goal

At the moment, the models don't output the data in a usable format (although you could extract it from the checkpoint files). Going forward we want to be able to download the model predictions. 

Initially this would involve downloading the Sea Ice Concentration in a netCDF format, but potentially add Zarr or numpy as alternatives.

Consider whether it would be useful to specify the dates to download or whether to just download the whole test period.

### What does good look like?

- [x] SIC data can be downloaded from model predictions

### Related context

_No response_

sylvesterkaczmarek:
## Summary

Add an opt-in path for exporting model predictions from `imp evaluate` in a usable NetCDF format.

## Changes

- add `--save-predictions PATH.nc` to `imp evaluate`;
- stream prediction batches directly to disk rather than retaining the full test period in memory;
- export `forecast_reference_time`, `lead_time`, `valid_time`, latitude and longitude coordinates;
- export every configured prediction target as a NetCDF variable;
- convert model outputs from the target dataset normalisation back to source units before writing;
- add CF metadata for sea-ice concentration (`ice_conc` / `sea_ice_area_fraction`);
- use the existing `data.split.test` ranges to control which dates are exported;
- document the workflow and single-process export requirement.

The writer is configured as a no-op evaluation callback by default, so existing evaluation behaviour is unchanged unless `--save-predictions` is supplied.

## Testing

- NetCDF creation and incremental multi-batch writes;
- denormalisation of prediction values;
- forecast-reference, lead and valid-time coordinates;
- latitude/longitude grid coordinates and SIC metadata;
- disabled callback behaviour;
- explicit rejection of multi-process export;
- CLI wiring and output-path resolution.

Closes #507.


IFenton (approved):
This looks good, and works nicely. I've altered the functionality a bit, so that the predictions are saved in the wandb run folder, rather than the user having to specify the location each time. But apart from that I'm happy to merge

sylvesterkaczmarek:
Thanks Ian. Good to see this merged with the run-directory integration, masks and ground-truth export included, and the full Python matrix green.

PR #514, opened 2026-08-27
Add climatology model 

AoifeHughes:
- Climatology model returns the `climatology` input dataset
- `climatology` dataset is created on-the-fly in the data module from the `target` dataset on first use
- Leap-year handling:
  - treat Feb 29th as its own date, falling back to using Feb 28th if there aren't enough data points.


sylvesterkaczmarek (commented):
Following up on my earlier comments: I like the overall climatology integration, but I still think the baseline should use a training-only day-of-year climatology rather than monthly means. The first-batch contamination of the climatology metric collection also still looks worth fixing so the baseline scores remain independent of the model predictions.

sylvesterkaczmarek (approved):
Re-reviewed current `006089e7`. The two issues I raised are resolved: the climatology is now a 366-day calendar-day table derived only from configured training periods, and the lazy climatology metric copy is reset before its first update so model-batch state cannot contaminate the baseline. The current test and lint matrix is green. No remaining blocker from me.

icenet_mp/data/common_data_module.py
```diff
@@ -175,6 +183,107 @@ def target_variable_indices(self) -> list[int]:
...
+                    table[index] = table[FEBRUARY_28_INDEX]
+                    continue
+                msg = (
+                    f"Cannot build climatology: calendar day {label} has no available "
+                    f"dates in the averaging period ({min(period_dates)} to "
+                    f"{max(period_dates)}). Check the configured training periods "
+                    "against the available data range."
+                )
+                raise ValueError(msg)
+            table[index] = target.get_tchw(day_dates).astype(np.float64).mean(axis=0)
```
Shaerdan, line 240: np.mean doesnt filter NaN cells. The SSMIS ice_conc store has no NaN in it (land is current valued as 0), so this does not affect the current pipeline, but if anywhere in the pipeline passes data gaps with NaN, they would silently corrupt the climatology. I suggest np.nanmean maybe a better replacement:

            table[index] = np.nanmean(
                target.get_tchw(day_dates).astype(np.float64), axis=0
            )

icenet_mp/models/climatology.py
```diff
@@ -0,0 +1,40 @@
...
+
+        Args:
+            target_variable_indices: The indices of the target variables.
+            kwargs: Additional arguments passed to the base model.
+
+        """
+        super().__init__(**kwargs)
+        self.automatic_optimization = False
+        self.model = nn.Identity()
+        self.variable_indices = target_variable_indices
```
Shaerdan, line 27: none binding suggestion: the attribute self.variable_indices is not used, could delete it.

Shaerdan (approved):
LGTM, the np.nanmean to replace np.mean suggestion is just a double safety thing, non-binding, i think it would be useful to make that one line change just in case some future dataset has NaN in them. But feel free to merge with or without it. Approved

PR #518, opened 2026-09-01
437 implement ddim

marianovitasari20:
**DDIM (Denoising Diffusion Implicit Models)** is a faster sampling method for a diffusion model. It does not require changing how the DDPM is trained. Instead, it changes how we run the reverse-diffusion process whenever we sample (i.e. validation, test, predict).

For example, say the **DDPM** has already been trained using 1000 diffusion timesteps. Normally, DDPM sampling would run the neural network at all 1000 timesteps:

> **noise → 999 → 998 → … → 1 → 0 → forecast**

With **DDIM**, we can skip most of those timesteps. For example, with `ddim_steps=50`, we pick 50 evenly-spaced timesteps out of the original 1000:

> **noise → 999 → 979 → 959 → … → 20 → 0 → forecast**

So we only need **50 network evaluations instead of 1000**, while reusing the same trained DDPM checkpoint. This speeds up every code path that rolls out the reverse process (validation, test, predict), by roughly 20×.

**`DDIMProcessor`** (`icenet_mp/models/processors/ddim.py`) is inherited from `DDPMProcessor` and only overrides `_run_reverse_diffusion` to implement DDIM sampling. The training loop, loss function, noise schedule, and v-prediction network are inherited unchanged, so an existing DDPM checkpoint can be sampled with DDIM without retraining.

**Denoising Diffusion Implicit Models (DDIM) paper**: https://arxiv.org/abs/2010.02502

<details>
<summary><strong>DDPM vs. DDIM (50 sampling timesteps) runtime, trained with 1,000 timesteps, using 2 GPUs on JASMIN and the full South dataset:</strong></summary>

* **DDPM:** 24+ hours (crashed after 24 hours) — https://wandb.ai/turing-seaice/train/runs/frm8qq30
* **DDIM:** 15h 22m 24s — https://wandb.ai/turing-seaice/train/runs/2ctrzd7z

</details>


sylvesterkaczmarek (commented):
re my earlier review, the DDIM formulation itself looks consistent with the existing v-prediction path. My remaining concerns are regression coverage for the sampler and validating the stochastic boundary, particularly preventing eta > 1 from silently producing an invalid variance update.

sylvesterkaczmarek (approved):
Re-reviewed the current head. The DDIM sampler now has direct coverage for the deterministic eta=0 path, the full-step boundary, and its timestep schedule, and eta is explicitly constrained to [0, 1], so the invalid stochastic variance case I raised cannot be silently clamped through. The implementation remains consistent with the existing v-prediction training path. No remaining blocker from me.

sylvesterkaczmarek (approved):
Rechecked current `1e15dce1` against the live PR diff. The PR-specific changes are still the DDIM implementation/config plus checkpoint override plumbing; the eta bound and deterministic/full-step regression coverage from my earlier review remain intact. The current Python 3.11, 3.12, 3.13 and 3.14 test matrix plus lint/build are green. No new blocker from me.

sylvesterkaczmarek (changes requested):
The new metric filter is broader than the problem it is fixing. For every multi-channel encoder it keeps only `mae` and `rmse`, which silently removes `ssim`; however `SSIMPerForecastDay` explicitly supports multi-channel tensors (channel-wise filtering/averaging), and the metrics guide documents SSIM as a general-purpose continuous-field metric rather than a sea-ice-specific one.

Channel count is also not a reliable proxy for dataset semantics: a single-channel non-SIC dataset would still inherit all SIC-only metrics. Could this filter the explicitly sea-ice-specific metrics instead (or use dataset/metric applicability), preserving generic metrics such as SSIM, and add a regression for the expected metric set on a multi-channel non-SIC EncoderStage?

marianovitasari20:
As per our last meeting on Monday, this PR is pending, waiting for the team’s decision. I’m in favour of handling it via the model service, as per the changes I made in the previous commits above. But if anyone has a better idea, please let me know. Maybe we can discuss this again in the next meeting, as I haven’t heard any updates on it yet.

marianovitasari20:
And please don’t update main into the branch yet. Whenever I update it, sometimes it breaks the functionality. It might be better to do it once we’ve agreed on the approach, so we don’t have to do it multiple times.

PR #534, opened 2026-09-08
Prevent validation leakage in normalisation statistics

### LINKED ISSUE #532: Normalisation statistics may include validation data in standard dataset splits
### Problem / Goal

The standard full-data split trains through `2021-12-31`, validates on `2022-01-01` to `2023-12-31`, tests on 2024, and reserves 2025 as the holdout set.

However, the full datasets span through 2025, and `SingleDataset.normalise()` uses the underlying Anemoi dataset-level `minimum` / `maximum` statistics.

For datasets covering 20+ years, Anemoi's default statistics period excludes the final three years. For a 1979-2025 dataset, that appears to mean statistics can include data through 2022. Anemoi date subsets also forward the statistics of the underlying dataset rather than recomputing them for the selected date range.

If so, 2022 validation observations influence the normalisation applied during training. This would introduce train/validation leakage and could affect validation loss, early stopping, hyperparameter selection, and model comparisons.

### What does good look like?

- [ ] Confirm whether the standard full datasets' stored normalisation statistics include any validation/test/holdout dates.
- [ ] Ensure statistics used for model training are based only on training-period data (currently no later than `2021-12-31` for the full split).
- [ ] Ensure validation, test, and predict/holdout dates cannot influence training normalisation.
- [ ] Add a regression/config test that detects overlap between the statistics period and non-training splits.

### Related context

Relevant IceNet-MP files:
- `icenet_mp/config/data/split/full_dataset.yaml`
- `icenet_mp/config/data/full_south.yaml`
- `icenet_mp/data/single_dataset.py`

Anemoi statistics behaviour:
https://github.com/ecmwf/anemoi-datasets/blob/main/docs/building/statistics.rst

sylvesterkaczmarek:
## Summary

- cap normalisation statistics at the training split boundary for the six full North/South ARGO, OSI SAF, and ERA5 dataset recipes
- add a regression test that requires the statistics cutoff to match the training cutoff and remain before validation, test, and holdout periods

## Why

The standard full-data split trains through 2021-12-31, while validation starts on 2022-01-01. Without an explicit statistics cutoff, Anemoi's default statistics window can include 2022 data, allowing validation observations to influence the normalisation used during training.

This change makes the intended split boundary explicit in the dataset recipes rather than changing generic dataset-loading behaviour.

## Validation

- full test suite: 1,912 passed
- targeted data-config tests: 9 passed
- pre-commit: all checks passed
- mypy: no issues in 250 source files
- verified Anemoi resolves the statistics end date to 2021-12-31 for all six affected datasets

Existing Zarr datasets will need their statistics regenerated or rebuilt for the new cutoff to take effect.

Fixes #532


jemrobinson (approved):
LGTM

